Issue #69, opened 2015-01-28
Add ESRI Map Display in Detail View


daveegrant:
Begin adding geospatial view of data using ESRI Maps JS API.  This ticket is to add a map display showing the location of a document that contains geospatial data.

dmcassel:
Do we want ESRI as the default mapping approach?

daveegrant:
After discussions withe Dave Cassel, we will include the ESRI Maps integration as an optional inclusion for the generator.  The default will be to NOT include the ESRI Maps integration code.

PR #70, opened 2015-01-29
#69 - Added ESRI Maps Integration

daveegrant:
Added ability to optionally install ESRI Maps integration.

### DIFF
--- a/app/templates/ui/app/app.js
+++ b/app/templates/ui/app/app.js
@@ -12,6 +12,7 @@ angular.module('sample', [
   'sample.search',
   'sample.common',
   'sample.detail',
+  'sample.esriMap',
   'sample.create'
 ])
   .config(['$routeProvider', '$locationProvider', 'mlMapsProvider', function ($routeProvider, $locationProvider, mlMapsProvider) {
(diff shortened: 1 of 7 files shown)


dmcassel:
I haven't had a chance to review this and probably won't for a couple weeks. Anybody else up for it?

grtjn:
@daveegrant could you append a commit to switch around detail.html and detail-no-esri.html. Make it detail.html (the default), and detail-esri.html. Maybe also comment the script tags for esri to save a bit on load time..

Otherwise I'd say, accept if it doesn't break anything. I will try to take a look on short notice..

grtjn:
I was rethinking my comment above. It would be supercool if it were possible to isolate esri stuff more as a plugin, or maybe some module that could be installed at will instead of just at initialization. Not sure slush could cater installing extra modules in a later stage, and not sure slush is the most appropriate tool for this either. Making a bower component out of it might make sense as well..

Having said that, doing it like this works good enough for now I guess.

And regarding my remark of making non-esri default: I didn't realize that it was done such that slush will by default remove esri stuff. It might have been nicer if it was done the other way around, but I guess inserting lines is more difficult than getting rid of them.

I'll have a quick run later this weekend to see if it runs, and merge the PR if it does..

grtjn:
Tested both with and without esri. Apart from the jsonEplorer issue, all worked well..

Issue #79, opened 2015-04-24
Detail view is not showing data


daveegrant:
After generating an app using slush, the detail view of a search result is not showing the data.  The new version of the json-explorer directive should use `<data>` as the attribute to identify the data rather than `<json-data>`.

grtjn:
One reason for this to keep turning up is that we don't have a version-specific dependency on json-explorer in bower.json. Last time I looked into that, I failed to change that. Maybe you have time to take a look at that as well?

daveegrant:
Added dependency on commit # for json-explorer since version details are not updated on release.

grtjn:
Fixed

PR #80, opened 2015-04-24
Fixed Display of Detail Data

daveegrant:
Updated use of json-explorer directive to properly display data.  Fixes issue #79 .

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -10,7 +10,7 @@
     "jquery": "~2.0.3",
     "ckeditor": "4.3.3",
     "ng-ckeditor": "latest",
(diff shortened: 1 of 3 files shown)


grtjn:
One reason for this to keep turning up is that we don't have a version-specific dependency on json-explorer in bower.json. Last time I looked into that, I failed to change that. Maybe you have time to take a look at that as well?

grtjn:
Looks like json-explorer is not releasing with versions properly. I'd be fine pointing to commits, that seems to be supported by Bower..

daveegrant:
Added dependency on commit.

PR #81, opened 2015-05-01
Added Integration with Google Maps

daveegrant:
Updates for Ticket #9 to add geospatial integration.  Based on feedback, chose to use Google Map to provide a search map and to update the detail-view to show a small map.  Updated slush file to prompt user to select map type: [none, google, esri].


grtjn:
Interestingly, this uses angular-google-maps. I have been using angular-ui-map a lot over the past year. Which is better you think? I also have a pretty advanced search-map directive, which behaves more like a drop-in on top of facets. I think it is worth taking a look at that code and compare things a bit. I'll send a link offline to a project using this..

grtjn:
I was also thinking about what if you would want to add esri/google afterwards. I think it would be nice if there was a small wiki page that sums up what you would need for that..

PR #82, opened 2015-05-01
added ml-lodlive to bower and included in index.html

withjam:
Now that ml-lodlive exists and is registered with Bower, Sid suggested it be included in the slush generator.  My updates simply add it to bower dependencies and include required resources in the index page.  Since it is designed to display triples data it didn't make sense to me to try to render anything without knowing what triples data it might be pointed at.  I also took the opportunity to make sure main.css appeared last in the list of css files so that it always wins out.

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -16,7 +16,8 @@
     "highlightjs": "8.0.0",
     "ml-utils": "withjam/ml-utils",
     "ml-search-ng": "~0.0.6",
-    "lodash": "~2.4.1"
+    "lodash": "~2.4.1",
+    "ml-lodlive": "latest"
   },
   "devDependencies": {
     "angular-mocks": "1.2.14",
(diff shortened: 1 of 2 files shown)


grtjn:
It would also be nice if you could provide a reference somewhere (inside detail.html, and detail-ctrl.js maybe) about how you would use this? Or maybe do it like esri/google integration and have a detail-lodlive.html and such that shows it in action. A small wiki page that sums up the details would be good as well.

It shouldn't be difficult to provide a snippet of triples you could load with MLCP, maybe something about colors? Then you could really show it in action.

Last thought: you could make it a slush option like google/esri. I think you can mimic what Dave did for that, so you won't be loading lodlive if you are not using it..

grtjn:
Oh, one other last thing: can you change 'latest' into a specific version? Just to protect against breaking changes. It is my intention to update bower.json periodically..

Issue #97, opened 2015-07-30
check for updates when generator is run


paxtonhare:
Check against github to see if any updates are available when generating a new template.

paxtonhare:
merged

PR #100, opened 2015-07-30
adding a check for latest version of the generator. if a newer versio…

paxtonhare:
…n is found a warning with instructions is printed.


grtjn:
Fixes #97

Issue #102, opened 2015-07-31
rename sample to app


paxtonhare:
No one wants to create a "real" application from this template and use the name sample.

Issue #77, opened 2015-04-14
Upgrade bower dependencies to latest


grtjn:
Lots of JS dependencies have advanced a lot. Bootstrap, Angular, but also ml-search-ng for instance. I recommend upgrading to the latest as much as possible.

grtjn:
PR from Paxton improves to large extend, but there are still newer versions of both bower and npm modules available.

grtjn:
Fixed

PR #112, opened 2015-08-13
#102 - renamed sample to app

paxtonhare:
- updated bower dependencies to latest
- moved route config into a separate file app/route/routes.js
- fixed a few issues in the gulp process for building and watching files
- renamed the node side app.js to node-app.js


grtjn:
Fixes (a.o.) #102

grtjn:
Also covers #77

Issue #103, opened 2015-08-04
Ports Specified Via Prompts


hunterwilliams:
Add questions during the initial creation to specify the app ports. It's two steps of manual changes that would be removed.

This would be especially helpful in cases where there may be multiple slush generated projects on a single MarkLogic development environment at once.

grtjn:
Bonus points for auto suggested ports.

Would be a useful addition to Roxy as well, though you cannot ask for user input with Roxy, when run headless from gulp/slush..

hunterwilliams:
This has been merged in. #151 is a continuation on this.

PR #113, opened 2015-08-18
Allow ports to be specified via prompts. #103

hunterwilliams:
Regards issue #103. It has the current defaults set as suggested ports.

One possible change that may be worth adding is some indication in the prepatched files that those values are going to be replaced. Also someone may want to double check or replace my find/replace with better regex to be number agnostic

### DIFF
--- a/slushfile.js
+++ b/slushfile.js
@@ -170,7 +170,7 @@ function runRoxy(config) {
 }
 
 // Make some changes to Roxy's deploy/build.properties file for the out-of-the-box application
-function configRoxy() {
+function configRoxy(appPort,xccPort) {
   console.log('Configuring Roxy');
 
   try {
@@ -179,6 +179,10 @@ function configRoxy() {
 
     // set the authentication-method property to digestbasic
     properties = properties.replace(/^authentication\-method=digest/m, 'authentication-method=digestbasic');
+
+    //set the ports
+    properties = properties.replace(/^app\-port=\b\d*\b/m, 'app-port=' + appPort);
+    properties = properties.replace(/^xcc\-port=\b\d*\b/m, 'xcc-port=' + xccPort);
 
     fs.writeFileSync('deploy/build.properties', properties);
   } catch (e) {
@@ -206,7 +210,7 @@ function configRoxy() {
 
 }
 
-gulp.task('default', ['init', 'configEsri'], function(done) {
+gulp.task('default', ['init', 'configGulp', 'configNodeExService', 'configEsri'], function(done) {
   gulp.src(['./bower.json', './package.json'])
    .pipe(install());
 });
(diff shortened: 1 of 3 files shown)


grtjn:
looks good at first glance!

grtjn:
Two additional thoughts:

- would be nice if you could specify ml-host as well. I typically run MarkLogic in a vm, as well as plenty others..
- would be nice if instead of doing a search/replace on two places to have to do it in one place only, and have both roxy and gulp/node read the same config. Is it possible to read the deploy properties files from node/gulp?

hunterwilliams:
* The host to specify would be easy. Would that just be for a local properties?
* I think there is a command in roxy to export the properties via json but I wasn't able to do it. It may not be in master yet? Or I was doing it incorrectly as I don't use roxy directly often

grtjn:
Just for local, good point. Well, you should put those port number in local then as well. That will be different for each env just as well..

Let me have a look at reading Roxy properties as JSON..

patrickmcelwee:
+1 for prompts, particularly since the app-port is set both in build.properties and the gulpfile (I wonder if there is a way to DRY that up). 

One suggestion: Given that most new apps are probably running against MarkLogic 8, shouldn't we default to port 8000 for XCC (and set install-xcc to false). That would reduce the number of choices to be made by 1.

grtjn:
You need the extra xcc for mlcp. Latest allows targeting a db, but only few have it installed..

patrickmcelwee:
Makes sense. We can revisit in a few years ...

ryanjdew:
Looks like this is all in working order. Any objections to merging, and perhaps opening an additional ticket, or tickets, for Geert's suggestions of config consolidation and specifying host?

paxtonhare:
I just tested this locally. @hunterwilliams it looks like you missed the port settings in routes.js. Those are used for authentication. 

```javascript
var options = {
  appPort: process.env.APP_PORT || 9070,
  mlHost: process.env.ML_HOST || 'localhost',
  mlPort: process.env.ML_PORT || '8040',
  defaultUser: 'admin',
  defaultPass: 'admin'
};
```

hunterwilliams:
Ah lovely. I'll fix that later on and get the conflicts out of the way

hunterwilliams:
I've added routes.js. I also found out that the gulpfile in the template was not properly adhering to gulp.config.js. I believe I've properly fixed it. Should be good to go.

grtjn:
Just thinking if we need to be able to override ports/host from cmd-line as well, if someone added Roxy overrides with local.properties..

hunterwilliams:
Not exactly sure I know what you're saying. Do you mean allow altering once the app has been generated via a command?

grtjn:
Yes, I am talking about overriding the ports/host after generating the app. You are replacing hard-coded values with other ones, and that is okay. But I typically override all with a local.properties and a parametrized call to gulp server (with the old version I mean).

local.properties:
```
local-server=ml8-ml1
app-port=8123
```

running ui:
```
gulp server --ml-host=ml8-ml1 --ml-port=8123 --app-port=9123
```

You want to allow people to easily override local settings, since that differs for everyone. If we allow the new `gulp serve-dev` to take the same params, then I'd already be happy. If we can find a way to read local.properties from gulp to use that, maybe even better?

slushfile.js
```diff
@@ -238,6 +242,42 @@ gulp.task('configEsri', ['init'], function(done) {
...
+});
+
+gulp.task('configRoutes', ['init'], function(done) {
+
+  try {
+
+    var routes = fs.readFileSync('node-server/routes.js', { encoding: 'utf8' });
+
+    //set the ports
+    routes = routes.replace(/\bprocess.env.APP_PORT \|\| 9070/m, 'process.env.APP_PORT || \'' + settings.nodePort + '\'');
```
paxtonhare: One final change requested @hunterwilliams .

Rather than replacing the lines in routes.js, could you just modify routes.js to read like proxy.js:

```javascript
var config = require('../gulp.config')();

var options = {
  appPort: process.env.APP_PORT || config.defaultPort,
  mlHost: process.env.ML_HOST || config.marklogic.host,
  mlPort: process.env.ML_PORT || config.marklogic.port,
  defaultUser: config.marklogic.user,
  defaultPass: config.marklogic.password
};
```

Then you don't need to replace anything. It will just work because of your configGulp task. This is my bad for not making them the same when I refactored the code base.

joemfb:
#129 was merged, so adding this comment here:

I'd like to see this also set the ports for the init scripts in #129. Maybe that should be a separate issue?

hunterwilliams:
@joemfb Would the port be the same?  

@paxtonhare will fix that

@grtjn After the change Paxton suggested is that necessary? Also I think maybe I should throw any more functionality in a separate pull request as it's a feature on top of this.

joemfb:
I think the ports should be the same.

hunterwilliams:
Awesome I'll make it change that file. Wasn't sure if the service port would be different by default since I saw it was specified as port 80.

grtjn:
Command-line params is a separate ticket now: #135 
Single config for ml-settings also: #151

We can always open another ticket and PR..

hunterwilliams:
This is ready to merge again.

joemfb:
Looks good!

Can you do an interactive rebase and squash this down to one commit?

hunterwilliams:
Down to one commit

PR #115, opened 2015-08-20
Replace Roxy generator questions with flags

patrickmcelwee:
hybrid apps no longer work and mvc apps are quasi-deprecated in Roxy

Also, in the generator, suggest the app-name provided by the user to
slush instead of cwd, which is probably not relevant.

Also includes general edits to the README, to add links and clean-up
formatting / flow a bit


paxtonhare:
@patrickmcelwee, I have one suggestion. Since the whole slush process is so verbose I tend to not read stuff. Could you make the warning about unsupported flags more prominent?

Here's what I see now:
```
[09:09:11] Starting 'marklogic-node:checkForUpdates'...
checking for latest version
[09:09:11] Finished 'marklogic-node:checkForUpdates' after 617 ms
[09:09:11] Starting 'marklogic-node:init'...
branchy=dev is not a supported flag and has been ignored. You can specify appType=<appType> to specify Roxy appType or branch=<branch> to specify a Roxy branch
[?] Name for the app? (blah)
```

Checkout what I did in gulpfile.js for the "new version available stuff:

```javascript
gutil.log(chalk.red('Warning: gulp version mismatch:'));
```

It's on line 100 of gulpfile.js

patrickmcelwee:
Good idea ... I just squashed that into my former commit (used `process.stdout.write(message.red);`, which was a style already in use in the slushfile)

PR #116, opened 2015-08-20
Auto-add local.properties to gitignore

patrickmcelwee:
to be consistent with how that property used in Roxy (machine-specific settings)


paxtonhare:
I accepted @hunterwilliams pull instead.

Issue #114, opened 2015-08-20
.gitignore dist folder


hunterwilliams:
slush-marklogic-node/app/templates/_gitignore

As a developer, I generally do not commit the **dist** folder. It's built every run. Is there any reason to exclude it from the default gitignore file?

grtjn:
+1, and while at it, these qualify as well I'd say:

```
deploy/local.properties
ui/styles/*.css
dist
*.pid
.idea
*.sublime-project
*.sublime-workspace
```

paxtonhare:
Make sure to be careful when putting folders in there. If you only want the top level dist then do:

/dist

In some projects I had a dist folder under some other folders and they also got ignored. oopsie.

PR #117, opened 2015-08-21
Added multiple folders/files to ignore

hunterwilliams:
Some of @grtjn 's suggestions as well as dist from my issue #114 also similar to PR #116 

Explanation of files added

* ui/styles/*.css - compiled css
* local.properties - see #116 
* /dist - see #114 - ignored the way @paxtonhare suggested
* *.pid - gulp uses locally each run
* *.idea/ - http://stackoverflow.com/questions/11968531/what-to-gitignore-from-the-idea-folder
* *.sublime-project - local sublime settings/think this is more of a choice than required
* *.sublime-workspace - open files. Definitely want gone
* .tmp/ - *please verify* but think it's used during build process and unnecessary

### DIFF
--- a/app/templates/_gitignore
+++ b/app/templates/_gitignore
@@ -1,4 +1,12 @@
 node_modules
 bower_components
 ui/fonts
+ui/styles/*.css
+deploy/local.properties
+/dist
 .DS_Store
+*.pid
+.idea/
+*.sublime-project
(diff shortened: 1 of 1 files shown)


grtjn:
"ignored Paxton" I think you mean you took his advice.. :)

@paxtonhare what do you think of the PR?

PR #123, opened 2015-08-26
Similar items

bluetorch:
show similar items (from extsimilar) in sidebar

### DIFF
--- a/app/templates/ui/app/detail/similar-directive.js
+++ b/app/templates/ui/app/detail/similar-directive.js
@@ -1,0 +1,25 @@
+(function () {
+
+  'use strict';
+
+  angular.module('app.similar')
+    .directive('mlSimilar', mlSimilar);
+
+  mlSimilar.$inject = ['MLRest'];
+
+  function mlSimilar(mlRest) {
+    return {
+      restrict: 'E',
+      templateUrl: 'app/detail/similar-directive.html',
+      scope: { uri: '@', limit: '@' },
+      link: function(scope, iElement, iAttrs, ctrl) {
+        mlRest.extension('extsimilar', { method: 'GET', params: { 'rs:uri': scope.uri, 'rs:limit': scope.limit ? scope.limit : 10 }}).then(function(response) {
+          scope.similar = response.data.similar;
+        });
+      }
+    };
+
+
+  }
+
+}());
--- a/app/templates/ui/app/detail/detail-no-esri.html
+++ b/app/templates/ui/app/detail/detail-no-esri.html
@@ -1,4 +1,9 @@
 <div class="row detail">
-  <!-- display JSON data -->
-  <json-explorer data="ctrl.doc"></json-explorer>
+  <div class="col-sm-9">
+	<!-- display JSON data -->
+	<json-explorer data="ctrl.doc"></json-explorer>
+  </div>
+  <div class="col-sm-3">
+    <ml-similar uri="{{ctrl.uri}}" limit="5"></ml-similar>
+  </div>
 </div>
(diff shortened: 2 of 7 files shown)


app/templates/ui/app/detail/similar-directive.js
```diff
@@ -0,0 +1,32 @@
...
+
+  mlSimilar.$inject = ['MLRest'];
+
+  function mlSimilar(mlRest) {
+    return {
+      restrict: 'E',
+      templateUrl: 'app/detail/similar-directive.html',
+      scope: { uri: '@' },
+      link: function(scope, iElement, iAttrs, ctrl) {
+        mlRest.extension('extsimilar', { method: 'GET', params: { 'rs:uri': iAttrs.uri, 'rs:limit': 10 }}).then(function(response) {
```
joemfb: This should be `scope.uri` (so that the value get's parsed/evaluated) instead of referencing the attribute directly. Also, could you parameterize the limit value as well?

app/templates/ui/app/detail/similar-directive.js
```diff
@@ -0,0 +1,32 @@
+(function() {
+  'use strict';
+
+  angular.module('ml.similar', []);
```
joemfb: Should this be `app.similar`, for consistency with the other internal modules?

joemfb:
I think it looks good. I've added some suggestions inline.

bluetorch:
I have made the suggested updates to make the code more consistent with the entire project.

joemfb:
Looks good! One last thing; it looks like `app/templates/ui/app/route/routes.js` got inadvertently included in this PR, as it's just a whitespace change? Can that file be removed from the PR, or is there some additional change needed?

Besides that, I vote for a merge (not sure what the normal approval process is).

bluetorch:
Whitespace removed.  routes.js no longer is in the pull request.

joemfb:
Awesome!

Not sure if it's the practice here, but I normally ask that multiple commits be squashed down to one, unless there's a good reason to keep them separate. Here's a guide I like: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html.

bluetorch:
Is that possible if I have already pushed and created a pull request?  My git skills aren't ninja-esque, so what I would know how to do is to delete the remote branch, squash the commits, then push to a new branch (of the same name) and submit a new pull request.

Opinion?

joemfb:
No, it's much easier than that. From your PR branch:

    git rebase -i HEAD~4

Then, edit the last 3 commits in the list, replacing `pick` with `squash`. Git will merge those 3 commits onto the first commit, and present you with an editor for writing a new message. Once you've committed with the new message, run `git log` to check that you indeed have only one commit on the branch, then force push:

    git push origin similar-items -f

paxtonhare:
@bluetorch I noticed you used a new module 'app.similar'. With the new refactor we should be making a similar.module.js that defines the module first. then the directive just uses it. You can see the example in create, search, or details.

Making the module separate allows us to isolate testing and conditionally include the module.

Also, rather than including 'app.similar' in app.js, you should include it in the detail.module.js. This moves the dependency to the thing that truly needs it vs the whole app. Again, this helps with testing things in an isolated manner.

bluetorch:
I'll make those changes.

bluetorch:
Updates made.  Ready for review.

app/templates/ui/app/app.js
```diff
@@ -11,6 +11,7 @@
     'app.detail',
     'app.esriMap',
     'app.search',
+    'app.similar',
```
paxtonhare: Yep. @grtjn is correct.

bluetorch:
Very silly.  I made the change and didn't save that listing.  Fixed now.  :blush:

Issue #124, opened 2015-08-27
Add Negation Support


hunterwilliams:
As a user, I would like to be able to take advantage of the upcoming version of ml-search-ng's negated facets.

NOTE: Requires bower updates for ml-search-ng and ml-common-ng

hunterwilliams:
@paxtonhare merged it in.

PR #125, opened 2015-08-27
Allow negated facets to work

hunterwilliams:
#124 

Note this needs to be pulled in only once ml-search-ng 0.0.2 is released! However this PR should allow negated facets to work soon as that happens.

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -13,7 +13,7 @@
     "highlightjs": "~8.7.0",
     "jquery": "~2.1.4",
     "lodash": "~3.10.1",
-    "ml-search-ng": "~0.1.0",
+    "ml-search-ng": "~0.2.0",
     "ml-utils": "withjam/ml-utils",
     "ng-json-explorer": "8c2a0f9104",
     "angular-mocks": "~1.4.4"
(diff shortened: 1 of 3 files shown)


joemfb:
joemfb/ml-search-ng#v0.2.0 has been released.

paxtonhare:
Is there any way, other than the color, to tell if an applied facet is negated or not? Seems like it shows blue if it's positive and orange if it's negated.

joemfb:
That's a good point, the only difference between "active" facets and "negated" facets in `ml-chiclets` is the bootstrap `btn-primary` vs `btn-warning` classes. Their should definitely be some more meaningful classes in there.

Did you have something else in mind?

grtjn:
How about a simple minus in front of the facet name? That is how you would type a not-query in the search box too, right?

jenbreese:
I thought the negated ones would have the ban icon and the active ones would have the plus.

grtjn:
Nope, the only difference visible in ml-search-ng/ml-facets.html is btn-primary versus btn-warning..

grtjn:
Please also rebase..

joemfb:
This PR won't be able to address the default negated chiclets style, as that's bundled with ml-search-ng. I've opened a ticket there for addressing these issues: joemfb/ml-search-ng#71.

hunterwilliams:
This is ready to merge again.

joemfb:
The history's a little weird in this branch, with a merge commit. Can you do an interactive rebase and squash it down to one commit?

hunterwilliams:
Squashed

Issue #119, opened 2015-08-21
Separate User Service?


hunterwilliams:
I know this one was talked about. I'm wondering at what level of detail does this take place? Or would this be a MarkLogic User based service? There are currently items in the app for like profile & email. However I'm not sure if those would be considered standard. 

I know the login service is supposed to be separated but this also sounds like it may either cross functionality or would have a dependency on that.

grtjn:
Maybe user and login service are best bundled together? Makes sense to me at least. Granularity could be useful though. One might want to change login ui, login method, and user management separately in a slush project..

grtjn:
Good enough for the moment..

Issue #118, opened 2015-08-21
Split Login From Project


hunterwilliams:
As a developer, I'd like to be able to reuse the login component.

I've done some initial work here: https://github.com/hunterwilliams/slush-marklogic-node/commit/294aec01d69783cbb91544575db126920d3233c0

Is that how it would be envisioned the app would interact with it? Possibly passing an endpoint then a preference of get/post. I know there is some modal work as well, but I assume that would be mostly forward facing.

Edit: 

I know some projects also allow OAuth Login. If the MarkLogic login is to be moved out (logging in via MarkLogic) should we also think about making it more in line with OAuth Authentication or something more generic and having something like this in the app - https://github.com/DaftMonk/fullstack-demo/blob/master/client/components/auth/auth.service.js ? That combined with some other node code allows OAuthLogins which could be useful (or not).

grtjn:
Isolating login stuff into its own module is a good first step. OAuth could be an interesting consideration, but perhaps if would be able to split off login as a bower component, we could have an oauth-login one with the same interface that could replace this one to support OAuth against some configurable server (which doesn't have to be MarkLogic perhaps?)..

@ryanjdew is working on authentication, perhaps talk with him as well..

ryanjdew:
This lines up a lot with what I'm working on. @hunterwilliams, I'll pull in your commit to make sure we're aligned on structure/API. 

As for OAuth, I think that would be good for the future but I feel it is a bit of a distraction from our goal for the upcoming weeks. Maybe we should open a separate ticket specifically for OAuth?

grtjn:
I think this is sufficiently covered for the time being. Feel free to open a new ticket for OAuth, closing this one..

PR #127, opened 2015-08-27
Move login out as service and add httpintercept for login prompt

ryanjdew:
Moved the login logic to it's own module.
Added an http intercept to bring up the login prompt when a 401 is hit.
Allows for an easy switch between full login page or a modal login prompt.
Changed nodejs proxy to require logins for REST API GETs, to match the default required login for search that previously was enforced on the front-end by default.


paxtonhare:
@ryanjdew Is there a simple way to keep the logout button and user icon and still use the modal or fullscreen logins?

To test I commented out the <ml-user...> in root.html and that seemed to enable the login.service to do its magic.

But I didn't see a way to keep the logout button w/o hacking on the user directive or manually creating it myself.

I'm thinking this pull request might also need a small blurb in the README on how to swap it around.

ryanjdew:
@paxtonhare, I added the blurb in the README for setting the appropriate variable. 

I also fixed the top right corner to have a link to login instead of the login boxes when a login mode other than 'top-right' is set.

app/templates/ui/app/user/user-dir.html
```diff
@@ -1,21 +1,15 @@
...
-      <input type="text" placeholder="Username" class="form-control input-sm" ng-model="ctrl.username">
-    </div>
-    <div class="form-group">
-      <input type="password" placeholder="Password" class="form-control input-sm" ng-model="ctrl.password">
-    </div>
-    <button type="submit" class="btn btn-success btn-sm">Sign in</button>
-  </form>
+  <div class="navbar-form navbar-right" ng-if="!ctrl.userService.currentUser()" >
+    <login ng-if="ctrl.loginService.loginMode() === 'top-right'" mode="inline"></login>
+    <span ng-if="ctrl.loginService.loginMode() !== 'top-right'" class="profile-link"><a ng-click="ctrl.loginService.loginPrompt()">Login</a></span>
```
paxtonhare: If you change this:
<a ng-click="ctrl.loginService.loginPrompt()">Login</a>

to this:

<a href ng-click="ctrl.loginService.loginPrompt()">Login</a>

then the cursor will change to the pointer cursor vs the text cursor w/o you having to set a style.

paxtonhare:
When _loginMode === 'full'  I'm seeing some weirdness with failed logins. Each time I do a failed login my url gets crazier:

http://localhost:3000/login?state=root.login&params=%7B%22state%22:%22root.login%22,%22params%22:%22%7B%5C%22state%5C%22:%5C%22root.login%5C%22,%5C%22params%5C%22:%5C%22%7B%5C%5C%5C%22state%5C%5C%5C%22:%5C%5C%5C%22root.search%5C%5C%5C%22,%5C%5C%5C%22params%5C%5C%5C%22:%5C%5C%5C%22%7B%7D%5C%5C%5C%22%7D%5C%22%7D%22%7D

Also, there is no indication that my login failed for any of the modes ('full', 'modal', 'top-right'). I simply just see the form again.

paxtonhare:
The /create route isn't forcing an auth prompt. Instead it's just showing up as empty in the ui-view

ryanjdew:
@paxtonhare , I believe I've worked through the remaining issues with the login redirects and the create route not forcing authentication.

grtjn:
Related to #118 and #119

Issue #44, opened 2014-09-09
Add template code for running as service with forever


grtjn:
Involves init and startup script, and a bit of documentation about how to set it up..

grtjn:
Fixed

PR #129, opened 2015-08-31
Providing scripts/docs to run app as service

bluetorch:
#44


app/templates/README.mdown
```diff
@@ -24,6 +24,38 @@ Edit `gulp.config.js` to set your desired ports
...
+- `$ [sudo] npm install forever -g`
+  
+Next, push all source files to the appropriate server. The following assumes it was dropped under /space/projects/ in a folder called slush-app.live. Take these steps to install the services:
+  
+- cd /space/projects/slush-app.live
+- gulp build # this will create the ./dist/ folder with all the required assests and code
+- cd /etc
+- sudo ln -s /space/projects/slush-app.live/etc/{env} slush-app
+- cd /etc/init.d
+- sudo ln -s /space/projects/dmlc-finserv.live/etc/init.d/node-express-service slush-app
```
grtjn: I think this should read slush-app..

app/templates/etc/prod/conf.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# js entry point
+#SOURCE_FILE=boot.js
+
+# server script to be launched
+#SERVER_SCRIPT=./node-server/node-app.js
+
+# port at which node server should be listening
+#APP_PORT=9040
+APP_PORT=80
```
grtjn: Is it wise to pick 80 as default? Don't we typically put Apache HTTPd in between? Not sure it is really necessary, but we typically do on demo servers. That port will likely not be available on those at least..

joemfb:
I'd like to see this also set the ports for the init scripts in #129. Maybe that should be a separate issue?

grtjn:
What should set which ports?

joemfb:
Doh, I put this on the wrong issue. Trying again...

I'd like to see these ports set automatically in the "project generation" process, as in #113.

Thoughts?

grtjn:
It kind of makes sense to keep them in synch as a kind of default, but on the other hand, the ports are usually different for each env..

app/templates/README.mdown
bluetorch: You are correct.  It should.  I will fix it. I blame this on clipboard inheritance.

app/templates/etc/prod/conf.sh
bluetorch: I just checked all of my POC demo boxes.  All of them have httpd installed, and all of them have the service off for each runlevel.  Here is an example:
[bholt@energy ~]$ chkconfig --list httpd
httpd          	0:off	1:off	2:off	3:off	4:off	5:off	6:off

bluetorch:
I think it is a good idea to have the ports specified from #113 be set in the init script, but not in the conf.sh

app/templates/etc/prod/conf.sh
grtjn: Yes, httpd will be delivered by it as installed but not launched out of the box. But it is still being used a lot. All our team servers have it enabled, as well as all servers I touched.. ;-)

PR #130, opened 2015-09-01
Add snippet size control

mariannemyers:
Add a user control to toggle between snippet sizes of 'compact' and 'detailed'. #132

### DIFF
--- a/app/templates/ui/app/search/search-results.html
+++ b/app/templates/ui/app/search/search-results.html
@@ -1,6 +1,6 @@
 <div ng-repeat="result in results">
   <h4>
-    <a ui-sref="root.view({uri: result.uri})">{{ result.title || result.uri }}</a>
+    <a ui-sref="root.view({uri: result.uri})">{{ result.uri }}</a>
   </h4>
   <div class="matches">
     <div class="match" ng-repeat="match in result.matches">
--- a/app/templates/ui/app/search/search-results.html
+++ b/app/templates/ui/app/search/search-results.html
@@ -1,6 +1,6 @@
 <div ng-repeat="result in results">
   <h4>
-    <a ui-sref="root.view({uri: result.uri})">{{ result.uri }}</a>
+    <a ui-sref="root.view({uri: result.uri})">{{ result.title || result.uri }}</a>
   </h4>
   <div class="matches">
     <div class="match" ng-repeat="match in result.matches">
--- a/app/templates/ui/app/search/search.controller.js
+++ b/app/templates/ui/app/search/search.controller.js
@@ -22,5 +22,11 @@
     $scope.$watch(userService.currentUser, function(newValue) {
       ctrl.currentUser = newValue;
     });
+
+    $scope.setSnippet = function(type) {
+      mlSearch.setSnippet(type);
+      ctrl.search();
+    };
+
   }
 }());
--- a/app/templates/ui/app/search/search.controller.js
+++ b/app/templates/ui/app/search/search.controller.js
@@ -19,14 +19,13 @@
 
     ctrl.init();
 
-    $scope.$watch(userService.currentUser, function(newValue) {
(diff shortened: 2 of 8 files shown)


app/templates/ui/app/search/search-results.html
```diff
@@ -1,6 +1,6 @@
 <div ng-repeat="result in results">
   <h4>
-    <a ui-sref="root.view({uri: result.uri})">{{ result.uri }}</a>
+    <a ui-sref="root.view({uri: result.uri})">{{ result.title || result.uri }}</a>
```
paxtonhare: Since we don't provide a title by default I would just do .uri.

Unless we show them how to easily supply a title.

app/templates/ui/app/search/search.controller.js
```diff
@@ -22,5 +22,11 @@
     $scope.$watch(userService.currentUser, function(newValue) {
       ctrl.currentUser = newValue;
     });
+
+    $scope.setSnippet = function(type) {
```
paxtonhare: our convention has been to put functions on the ctrl variable:

ctrl.setSnippet = ....

see below for the html approach

app/templates/ui/app/search/search.html
```diff
@@ -12,10 +12,12 @@
...
 
     <div ng-if="ctrl.response.total > 0" class="results">
-
-      <pagination class="pagination-sm" ng-model="ctrl.page" ng-change="ctrl.search()" max-size="10" boundary-links="true" total-items="ctrl.response.total" items-per-page="ctrl.response['page-length']"></pagination>
-
-      <ml-metrics search="ctrl.response"></ml-metrics>
+      <div class="pagination-ctrls">
+        <pagination class="pagination pagination-sm" ng-model="ctrl.page" ng-change="ctrl.search()" max-size="10" boundary-links="true" total-items="ctrl.response.total" items-per-page="ctrl.response['page-length']">
+        </pagination>
+        <ml-snippet set-snippet="setSnippet(type)"></ml-snippet>
```
paxtonhare: We should be doing ctrl.setSnippet

app/templates/ui/app/search/search-results.html
grtjn: I like having the possibility of showing a title or label instead of uri. It is very common to want to not show the uri. I would not mind at all keeping this..

patrickmcelwee: I vote you show me ... erm ... them how to supply a title (maybe just a comment here, or maybe a recipe in the docs?)

mariannemyers: I've previously used transforms that pass through most (or all) of the
search results and add in a title that is appropriate for the dataset.  I
could write that up in more detail, or we could add a transform to this
application.  Opinions?

grtjn: Another way could be to just loop over response.search.results, and add title/label (I personally like `label` better by the way) before adding it to $scope or ctrl. Such label could be derived from metadata, extracted data, raw content, uri, or something else..

A nice default pattern could be to take the uri, strip away the path, and keep just the doc name as result.label..

joemfb: If that kind of pattern is going to be added, I would ask that it be added to the `ml-results` directive. Similar processing is already being done to add a `link` property to each result object: https://github.com/joemfb/ml-search-ng/blob/master/src/directives/ml-results.directive.js#L59

mariannemyers:
Given the level of discussion on generating the label/title, I propose we only use the PR for the snippet size control.  Let's further discuss how we should be generating the label, and if it should be part of ml-results as another task.

grtjn:
Makes sense. If we want to push part of functionality into ml-search, then this would have to wait for a release of that. Let's not do that (wait I mean, I am in favor of adding something to provide a label in ml-search)..

Issue #56, opened 2014-11-14
Add default favicon


grtjn:
A default favicon is missing, and the remark to put a favicon.ico in the root is no longer the recommended method. Add explicit links in index.html, and add a ML favicon as default. Makes it easier to find the appropriate place to put a favicon as well.

grtjn:
Fixed

PR #155, opened 2015-09-03
Adds a favicon and adds it to index.html

hunterwilliams:
This would complete #56 

Not exactly sure what the icon is originally from but I had it from another ML project.


joemfb:
Looks good!

One comment: according to https://stackoverflow.com/questions/10218178/necessary-to-add-link-tag-for-favicon-ico, `rel="shortcut icon"` is from an old version of IE; the new standard is just `rel="icon"`.

hunterwilliams:
I'm game but wondering... How far back do we support?

grtjn:
This site also mentions apple-touch-icon something: http://iconhandbook.co.uk/reference/examples/favicons/

I wouldn't mind that shortcut bit, its just 9 chars, and harming no one.. ;)

hunterwilliams:
@grtjn I suggest we avoid the apple-touch-icon. It opens a can of worms and then you need a bunch of different sizes etc. If we were pushing a mobile based app sure but I don't think we're doing that.


@joemfb Would it be okay to just leave it in for legacy purposes?

joemfb:
I'm not attached one way or another; just sharing what I had learned ;)

hunterwilliams:
<link rel="shortcut icon" href="http://owwgdpc6ll4w6eh4717xd1x4.wpengine.netdna-cdn.com/wp-content/mu-plugins/images/favicon.ico?cache=2">

That's what marklogic.com does. I've changed the favicon to use that one. Squashed the commit down to one.

grtjn:
Wow...

grtjn:
Odd, not seeing the update you commented. Forgot to push maybe?

hunterwilliams:
Sorry I squashed it down. Should have waited.  It's there though :)

grtjn:
Squashing it should not hide it from the diff. I am really not seeing it, just:

```
<link rel="shortcut icon" href="/images/favicon.ico" />
```

and a binary we no longer need..

hunterwilliams:
Sorry I meant I grabbed the binary from the cdn and used that. I don't think we should link to MarkLogic's cdn

grtjn:
Ah cool, even better!

grtjn:
Worked nicely for me, merging..

Issue #136, opened 2015-09-03
Search data only


grtjn:
Deploy content uploads dictionary, but dictionary appears in search results. Best to apply collection `data` (or similar) to sample-data, and isolate search options to that.

Also nice to strip away abs path:
```
./ml local mlcp import -input_file_path sample-data -output_uri_replace ".*/sample-data/,\'/sample-data/\'" -output_collections data
```
and:
```
  <search:additional-query>
    <cts:collection-query xmlns:cts="http://marklogic.com/cts">
      <cts:uri>data</cts:uri>
    </cts:collection-query>
  </search:additional-query>
```

hunterwilliams:
Instead could we throw the dictionary in its own collection and do a not query on it?

I feel like it may be a pain to make sure you always add the data collection to any document vs just fixing this one

grtjn:
The not-query that we have now is a bad practice, adding the dictionary to it won't make it better. The collection-query is a much better approach, and very easy to explain. You can also do lots of fun with collections like adding collections like `data/cars` and `data/houses`, and then doing a collection facet with `data/` prefix..

grtjn:
We could add that as a sample facet actually.. :)

PR #157, opened 2015-09-04
Added more sample docs

hunterwilliams:
If someone could verify the command that would be great. It should be correct but was having issues getting mlcp to work on my project properly.. #159


app/templates/README.mdown
```diff
@@ -71,26 +71,26 @@ Next, push all source files to the appropriate server. The following assumes it
...
 
 ## Sample Data
 
-The application comes with 25 JSON documents generated by json-generator.com.
+The application comes with 1300 JSON documents generated by json-generator.com.
 You can load them with
 (MLCP)[https://docs.marklogic.com/guide/ingestion/content-pump] like this:
 
-    ./ml local mlcp import -input_file_path sample-data -document_type json
+    ./ml local mlcp import -input_file_path sample-data -document_type json -input_compressed true
```
grtjn: While in here, can you make some more changes to this line?

- You don't need the `-document_type` param, MarkLogic will auto-detect from extension.
- You can save yourself typing `true`, just `-input_compressed` is enough.
- Could you please add `-output_collections data`, both a good practice, and useful for changing the additional-query to a better default
- Last but not least, we should apply doc-permissions to allow the app-user that gets generated automatically access as well: `output_permissions <app-name>-user:read,<app-name>-user:update`

Oooh, we can automate the <app-name> replacement. Put the above options in an `import-sample-data.options` file, and use @ml.app-name inside there! :)

grtjn: Regarding output_collections, see also #136..

hunterwilliams:
I'll tackle both at once. Sorta confused on the import-sample-data.options. Will that automatically be modified during the generation process and swap the app name to what we want? Also I'm assuming roxy or mlcp can use the options file?

grtjn:
Roxy does some magic to options files. It reads them, resolves those @ml. patterns, and appends the contents to the system-call it makes to run MLCP..

hunterwilliams:
So the file should read like this?? 

``
local mlcp import -input_file_path sample-data.zip -input_compressed -output_collections data -output_permissions @ml.app-name-user:read,@ml.app-name-user:update
``

Or does <@ml.app-name>-user ? Not sure how roxy does the replace

grtjn:
Close, but MLCP option files should have each argument on a separate line. The MLCP call would read:

```
./ml local mlcp -options_file import-sample-data.options
```

And the options file could contain:
```
import

-input_file_path
sample-data/

-input_compressed

-output_collections
data,data/people

-output_permissions
@ml.app-name-user:read,@ml.app-name-user:update
```

That should do it..

hunterwilliams:
I've added that. Think the search options change should be a separate PR actually. Similar documents doesn't read from search options and going to have to change creating documents to get the collection as well

grtjn:
Yep, separate PR for search options (and similar) makes sense. Good point about similar. Can you create a ticket for that?

grtjn:
PS: changes look great, haven't actually tested yet though..

ml:
Haha, whenever you use @ml, I get a notification ;) I guess I have to ignore the repo ...

grtjn:
Yes, I noticed. I'll try to use back-quotes around it as much as possible.. :)

grtjn:
Almost there.. `@ml.default-user` is both wrong, and does not work properly. Roxy mistakenly looks for `-option_file` instead of `-options_file`. It also inconveniently looks for it in the `deploy/` directory. And permissions work on roles, not users. So it should have been `@ml.app-role`..

Perhaps just put a placeholder there, and ask people in the README to edit the options..

grtjn:
Last but not least, I got the syntax wrong, last line should use just commas, no colons:

```
$app-role,read,$app-role,update
```

grtjn:
Sorry, not last. Not that I mind too much, but there are 3000 docs in the zip, not 1300. And can you also append this to the options file, to strip away the zip path from the uris:

```
-output_uri_replace
".*\.zip,''"
```
Note, that are two single quotes, in double ones..

grtjn:
Otherwise import working nicely, so after those fixes this should be good to go..

hunterwilliams:
All changes should be in now

hunterwilliams:
@grtjn see if this satisfies #169

grtjn:
Nice, getting 7 eyeColors. Definitely better. Now the show more has a purpose too.. ;)

grtjn:
I think I ran sufficient tests with this, merging..

Issue #110, opened 2015-08-12
Browser Caching Issues In Edge


hunterwilliams:
As a user, I'd like to get be able to get new data/login properly when using a Microsoft browser (Edge tested - IE likely affected as well). 

@daveegrant asked for help to check out a possible issue and working him found that browser cache seems to be an issue

See this: http://stackoverflow.com/questions/16971831/better-way-to-prevent-ie-cache-in-angularjs

Recreation:

Start server
Windows 10 Edge Browser - Login slush generated app. Logout
Stop server
Windows 10 Edge Browser- Login again. It works with the server off. See network logs. Reads something like

``
Name	Protocol	Method	Result	Content type	Received	Time	Initiator
http://your-server-here.com:8001/user/login?password=password&username=user	HTTP	GET	200	application/json	(from cache)	0 s	
``

daveegrant:
An alternate to the suggestion in the linked article is to add a noCache function that adds the response headers to prevent caching.  That function can be invoked in the proxy function and any other function where it makes sense (e.g. - /user/login).  The advantage of this approach is that requests for static resources will not include the no-cache response headers since it makes sense for them to be cached.

grtjn:
+1 for adding no-cache response header to POST/PUT/DELETE in middle-tier..

hunterwilliams:
So should we go ahead and do this on all calls? Or just some? 
Also I'm not sure if a combined js file is created with a hash each change or not - I know that's an option that some node projects use.

grtjn:
In middle-tier for any dynamic POST/PUT/DELETE (the login is a dynamic call). The hash is more about the static files, but could be useful as well.

hunterwilliams:
@daveegrant Can you verify I did this correctly? See #158

PR #158, opened 2015-09-04
No Cache Requests

hunterwilliams:
A bug with edge browser/IE based browsers caching requests - see #110 

Not sure this is the perfect solution but I didn't want to try and force all requests to be no-cache

### DIFF
--- a/app/templates/node-server/proxy.js
+++ b/app/templates/node-server/proxy.js
@@ -18,6 +18,7 @@ var options = {
 // ==================================
 // For any other GET request, proxy it on to MarkLogic.
 router.get('*', function(req, res) {
+  res.append('Cache-Control', 'no-cache');
   if (req.session.user === undefined) {
     res.send(401, 'Unauthorized');
   } else {
@@ -28,6 +29,7 @@ router.get('*', function(req, res) {
 });
 
 router.put('*', function(req, res) {
+  res.append('Cache-Control', 'no-cache');
   // For PUT requests, require authentication
   if (req.session.user === undefined) {
     res.send(401, 'Unauthorized');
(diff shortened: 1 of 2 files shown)


grtjn:
Maybe put that in a helper function, and then add a few more headers to be paranoid:

http://stackoverflow.com/a/4485194/918496

;-)

hunterwilliams:
I could sorta toss into a helper function however not sure about adding all of those headers.. Not sure if it affects anything though

grtjn:
Not sure how important they are. Perhaps just for legacy browsers? Then again, not a big deal either. It won't break anything.. :P

grtjn:
How about adding the helper function, and add those other headers as commented lines with a comment above them reading something like: For legacy browsers

hunterwilliams:
Done. May want to move into another file it if gets any more complicated

grtjn:
Hmm, I added a proxy rule for DELETE in PR #177 (see also #176). Think I should rebase that PR after this gets merged..

grtjn:
@hunterwilliams I think we can drop the noCache for /v1 GET (proxy.js). Otherwise runs fine on my end..

grtjn:
On second thought, maybe we should drop the noCache for MarkLogic entirely (e.g. don't apply to proxy.js). MarkLogic should be returning such headers itself, and my #177 PR makes sure those headers are forwarded..

grtjn:
On third thought, not really sure MarkLogic does properly.. :-/

hunterwilliams:
Unfortunately MarkLogic doesn't send the headers so your searches get cached...

grtjn:
Yes, indeed. Do we want GET calls to MarkLogic get cached or not? Most searches use POST I think. Maybe a noCache for GET does sound right after all?

hunterwilliams:
I'd assume we don't want anything to MarkLogic cached. Someone will eventually add some GET endpoint and then be hosed on IE and can't figure out why GET requests seem to fail

grtjn:
ok, no further changes required then..

Issue #161, opened 2015-09-04
Make Similar Docs Respect Search Options


hunterwilliams:
Or at least make it have similar options.

PR #162, opened 2015-09-04
Made it so that search options query on data collection, similar docs…

hunterwilliams:
#161 
#136 

Made it so that similar, search options, and people creation properly utilize the data collection.

### DIFF
--- a/app/templates/ui/app/create/create.controller.js
+++ b/app/templates/ui/app/create/create.controller.js
@@ -45,7 +45,8 @@
       mlRest.createDocument(ctrl.person, {
         format: 'json',
         directory: '/content/',
-        extension: '.json'
+        extension: '.json',
+        collection: ['data','data/people']
         // TODO: add read/update permissions here like this:
         // 'perm:sample-role': 'read',
         // 'perm:sample-role': 'update'
--- a/app/templates/rest-api/config/options/all.xml
+++ b/app/templates/rest-api/config/options/all.xml
@@ -4,20 +4,18 @@
 
   <!-- Limit all searches to this additional query -->
   <additional-query>
-    <cts:not-query xmlns:cts="http://marklogic.com/cts">
-      <cts:directory-query>
-        <cts:uri>/users/</cts:uri>
-      </cts:directory-query>
-    </cts:not-query>
+    <cts:collection-query xmlns:cts="http://marklogic.com/cts">
+      <cts:uri>data</cts:uri>
+    </cts:collection-query>
   </additional-query>
-  
+
   <!-- Full-text search options -->
   <term apply="term">
     <empty apply="all-results"/>
     <term-option>punctuation-insensitive</term-option>
     <term-option>stemmed</term-option>
   </term>
-  
+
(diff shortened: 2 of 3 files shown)


app/templates/ui/app/create/create.controller.js
```diff
@@ -44,7 +44,8 @@
       mlRest.createDocument(ctrl.person, {
         format: 'json',
         directory: '/content/',
-        extension: '.json'
+        extension: '.json',
+        collection: ['data','data/person']
```
grtjn: Was it data/person, or data/people?

hunterwilliams:
@grtjn Nice catch. I changed it on my sample project and thought I committed that...

joemfb:
Could the collection name be an optional parameter to `extsimilar`?

hunterwilliams:
Now should work! @joemfb 

If you don't pass a collection it just searches over all documents

grtjn:
You could also consider sending options name, pull up persisted options, blend in similar query, and do a search:search. But I could have a whack at that myself. Might wanna move it to an mlpm package as well..

grtjn:
search-utils can be used to grab persisted options, see example here: https://github.com/grtjn/ml-decorators/blob/master/facet-values-decorator.xqy#L14

hunterwilliams:
@grtjn my only concern / possible annoyance would be that you want to filter further than your general search options and just to change this you'd have to make another search options. I see the benefits but if a developer is going to really need custom search options for each type they're likely savvy enough to fix the directive for their particular use.

grtjn:
Not necessarily. You could make it such that you can also send along a string query, or a structured query which is more powerful than just a collection param. Then again, this really sounds like rebuilding /v1/search, which already takes all these params, and allows sending along override query options etc. Can't we just send a cts:similar-query to that somehow? Maybe I need to have a closer look at this..

hunterwilliams:
I mean you could yank the collections from the current document and use that in the search or it could possibly grab the all search options and do the similar query with those.

grtjn:
Interesting thought pulling collections from the doc in question itself. :)

/v1/search allows return similar, by the way. Bit awkward with binaries, but could be a sensible alternative..

hunterwilliams:
Going to just make default to the document's collections by default and remove the stuff from the directive. Going with the keep it simple approach after talking with @grtjn

grtjn:
Yes, we can work on a more advanced extsimilar as a separate component if we think that is useful. A version that would go much more towards /v1/search would be interesting. Should also handle binaries better.

Looking forward to your updates! ;)

hunterwilliams:
Ready to be verified and merged. May need to squash down commits

grtjn:
I'm happy, but let's hold off a little if anyone else has comments..

grtjn:
The details page has been touched in a different PR. I'm afraid that caused a merge conflict. Can you undo that detail page change, and then rebase?

hunterwilliams:
Rebased and updated

grtjn:
Perfect, runs smoothly for me..

PR #163, opened 2015-09-07
Fixes #160: added collection facet, added comments, removed verbose search: prefix

### LINKED ISSUE #160: Add a collection facet (with prefix?)
If we add collections to the sample-data anyhow, we could also add a collection facet by default. Simple, and looks fun..

grtjn:
#160


grtjn:
Merging now, allows #162 to stack on top of this..

PR #166, opened 2015-09-07
Fixed #144: fixed issue with add/remove tags on create page

### LINKED ISSUE #144: Cannot add tags (nor friends)
on Create page.. tags should work, but broken. Adding friends not available, but details show a friends array?

grtjn:
Looks like this small detail got overlooked in the major rework. Also added removeTags while in there.. #144


grtjn:
Merging to allow @jenbreese to check the UI on this..

Issue #143, opened 2015-09-03
Clicking search before login throws TypeError


grtjn:
Looks like the 401 is not handled elegantly, and things go wrong in search-ctrl or maybe even deeper down (in ml-search perhaps).

joemfb:
It looks like the interceptor is preventing `$http` from properly rejecting the 401: https://github.com/marklogic/slush-marklogic-node/blob/master/app/templates/ui/app/error/errorInterceptor.service.js#L14

grtjn:
@ryanjdew or @hunterwilliams You two have played with login, right? What do you make of this?

grtjn:
Related to #143?

hunterwilliams:
Good question it actually may be. I've got the fix for the other issue so I'll test this out to see if it still pops.

hunterwilliams:
Can confirm it appears entirely related to #141

grtjn:
Fixed with PR #168 as well..

PR #168, opened 2015-09-07
#141 PR - Login Service Changes/Fix

### LINKED ISSUE #141: Clicking create requires (second) login
After logging in for search, you need to login again if you open create.

hunterwilliams:
#141 targeted
#143 also fixed when adding search to routes.js

Let's discuss what I've got so far... It's going to require possibly some design changes further than what I've done so far


app/templates/ui/app/login/login.service.js
```diff
@@ -9,9 +9,10 @@
   function LoginService($http, $modal, $q, $rootScope, $state, $stateParams, messageBoardService) {
     var _loginMode = 'full'; // 'modal', 'top-right', or 'full'
     var _loginError;
-    var _isAuthenticated = false;
     var _toStateName;
     var _toStateParams;
+    var _isAuthenticated = isLoggedIn();
```
grtjn: Is it your intention to eval the isLoggedIn function right on the spot, or do you want to make isAuthenticated (way down) return a promise from that helper function? (it is doing first, but kinda guessing you want the latter?)

grtjn:
Are any further changes needed in user-service or such? You talked about cyclic dependencies, but this looks clean..

app/templates/ui/app/login/login.service.js
hunterwilliams: It's currently returning the promise I think. At least the value isn't there in time for the bottom. Ideally I don't want to wait for a promise.

hunterwilliams:
User-service already depends on login-service. That's why I moved over checking status vs making login-service depend on user-service

grtjn:
Yeah, see no reason why login should depend on user service. Looks good to me.. :)

hunterwilliams:
Ill start attempting to make the promise work properly the first go as right now it doesn't resolve in time. I'll also update the tests

app/templates/ui/app/login/login.service.spec.js
```diff
@@ -67,4 +67,60 @@
...
+    }));
+
+    it('should be defined', function () {
+      expect(service).to.be.defined;
+    });
+
+    it('should be logged in already', function() {
+      expect(service.isAuthenticated()).to.be.not.defined;
+      service.getAuthenticatedStatus().then(function(done) {
+        expect(done).to.be.eq(100);
```
hunterwilliams: having issues with this this line being ran no matter what...

app/templates/ui/app/login/login.service.js
```diff
@@ -117,20 +154,24 @@
...
-            deregisterLoginSuccess = null;
-            $state.go(next.name, nextParams);
+
+      if (routeIsProtected(next.name)) {
+        var auth = getAuthenticatedStatus();
+
+        if (angular.isFunction(auth.then)) {
+          auth.then(function() {
+            if (!isAuthenticated()) {
+              //this does NOT block requests in a timely fashion...
```
hunterwilliams: Got it working so the UI seems correct to the user however it still pushes out requests a bit fast. may need to tweak block route or the login interceptor?

grtjn:
Can @ryanjdew or someone else with more experience with auth in angular review this PR?

hunterwilliams:
Also please let me know what I need to do to fix part of it. See the code comments

hunterwilliams:
My last commit added root.search to protected routes. It never made it back into slush from my sample project apparently. :frowning: Solves issue of "early" requests....

app/templates/ui/app/login/login.service.spec.js
grtjn: does this still apply with your latest commit?

hunterwilliams:
After some more prodding think it's good. Ready to review & merge :)

grtjn:
Excellent, works good for me. Even when I try to open /profile directly..

PR #170, opened 2015-09-07
Fixed #135: handle app-port arg properly, only open debug port when debug flag

### LINKED ISSUE #135: gulp serve-dev ignores app-port
This line in gulpfile.js looks suspicious:

```
var port = process.env.PORT || config.defaultPort;
```
It ignores args['app-port'] (should be checked first), and looking at where it is used, not clear why code needs this separate var, while all other options come from config/options structures?

Also serve-dev always runs at 3000, unless you have a PORT env var. I'd rather have an extra debug-port or sync-port option instead, and maybe a default in the config somewhere..

grtjn:
#135

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -15,8 +15,6 @@ var _ = require('lodash');
 var $ = require('gulp-load-plugins')({lazy: true});
 /* jshint ignore:end */
 
-var port = process.env.PORT || config.defaultPort;
-
 /**
  * yargs variables can be passed in to alter the behavior, when present.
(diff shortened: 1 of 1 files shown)


grtjn:
Anyone like to review this?

patrickmcelwee:
Nice work. I tested this out locally, with and without debug (and debugBrk - wondering why you added that option?), with the various ways of setting ports and ml-host, and with 2 gulp processes. It all worked well for me.

grtjn:
You will have to ask Paxton about debugBrk.. ;-)

PR #172, opened 2015-09-08
Fixed #85: added robots.txt

### LINKED ISSUE #85: Add default robots.txt's
In both ui/app/ and src/ containing:

User-agent: *
Disallow: /

grtjn:
#85

### DIFF
--- a/app/templates/src/robots.txt
+++ b/app/templates/src/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /
+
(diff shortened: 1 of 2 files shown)


hunterwilliams:
Looks it should block the robots who observe robots.txt

grtjn:
Yes, would be nice if we could stop all with just a .txt, but this is at least something.. ;-)

grtjn:
Merging, any further comments/changes will have to go into a new ticket/PR.. ;-)

Issue #176, opened 2015-09-09
Proxy doesn't support DELETE requests


grtjn:
Not used by vanilla app at the moment, but something devs would expect to be implemented. Should be a trivial fix..

Issue #67, opened 2015-01-16
server.js should honor Content-Disposition and pass-through header


garyvidal:
When using rest extensions, the server.js should honor a request with Content-Disposition and forward it as a header back to application

To resolve issue 
```js
      //If the file is a download then set the header to passthrough
      if(response.headers["content-disposition"]) {
          res.header("Content-Disposition",response.headers["content-disposition"]);
      }  
```

grtjn:
I think I would vote for a more generic solution in which request and response are passed through entirely, maybe even as-is if possible.

I have successfully used the piping features of the 'request' module. Most essential part would be replacing the existing proxy function with:

```js
  function proxy(req, res) {
    var queryString = req.originalUrl.split('?')[1];
    var mlOptions = {
      method: req.method,
      uri: 'http://' + options.mlHost + ':' + options.mlPort + req.path + (queryString ? '?' + queryString : ''),
      headers: req.headers,
      auth: getAuth(options, req.session)
    };
    try {
      //console.log(req.body + '');
      req.pipe(request(mlOptions)).pipe(res);
    } catch (e) {
      console.log(e);
      res.status(500).send('internal error');
    }
  }
```

joemfb:
I like the proxying approach as well. Much nicer to avoid the low-level request wrangling the current approach requires

grtjn:
I heard there is a new node module that can do proxying: gulp-webserver. I think it is worth looking into..

withjam:
Also check out a module 'http-proxy'.  I have used it in a couple of projects and swapped it in OBI as well.  It's very easy to setup and lets you proxy websockets as well.  It looks like this:

var httpProxy = require('http-proxy');
var mlProxy = httpProxy.createProxyServer( { target: dbUrl } );

  function proxy(req, res) {
    req.headers.Authorization = getAuth(appConfig, req.session);
    mlProxy.web(req,res); // this handles the multipart parsing and everything
  }

grtjn:
So as a replacement for 'http' we have:

- request
- gulp-webserver
- http-proxy

Which makes most sense? I believe request was able to replace http entirely. Does that also apply to gulp-webserver and http-proxy? Can anyone tell?

joemfb:
I used `request` here: https://github.com/joemfb/mlpm-registry/blob/master/server.js#L67 (see the `proxyConfig` and `proxy` functions. I like that flexibility over just using a proxy implementation.

grtjn:
[...]

PR #177, opened 2015-09-09
Fixed #31,#67,#176: minor changes to proxying to smoothen things a bit

### LINKED ISSUE #31: Issue with proxying data to ML
The current server.js implementation is using a bodyParser:

    app.use(bodyParser.json());
    app.use(bodyParser.urlencoded({
      extended: true
    }));

These lines seem to interfere with POST/PUT of JSON and binary to MarkLogic. Passing through JSON has a weird effect that MarkLogic keeps reading request body until it hits the length limit. Binary doesn't come through either.

I had to disable above lines, or revert to the 'request' library, which allows direct piping of request and response. Unfortunately, both solutions cause deprecation warnings to be shown, and the latter looks like it is a bit slower, but that is bit difficult to prove..

grtjn:
#31, #67, #176

### DIFF
--- a/app/templates/node-server/proxy.js
+++ b/app/templates/node-server/proxy.js
@@ -56,6 +56,16 @@ router.post('*', function(req, res) {
   }
 });
 
+// (#176) Require authentication for DELETE requests
+router.delete('*', function(req, res) {
+  noCache(res);
+  if (req.session.user === undefined) {
+    res.send(401, 'Unauthorized');
+  } else {
+    proxy(req, res);
+  }
+});
+
 function getAuth(options, session) {
   var auth = null;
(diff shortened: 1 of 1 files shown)


grtjn:
Odd, ml-highcharts-ng tries to communicate with MarkLogic itself. If we put the bodyParser above the proxy it gets the necessary cookie, but with the bodyParser below the proxy it doesn't, and the call hang in Chrome, showing (pending), and MarkLogic ErrorLog reports socket issues..

grtjn:
Anyone good with bodyParser care to have a look?

grtjn:
Taking out the fix for #31, want to investigate that more closely. The other fixes are straight-forward..

PR #179, opened 2015-09-10
CreateController - Fixed lint error and made tag test work properly

hunterwilliams:
Saw issues during gulp test. Fixed now. Project could use a Jenkins CI or something #144

### DIFF
--- a/app/templates/ui/app/create/create.controller.js
+++ b/app/templates/ui/app/create/create.controller.js
@@ -46,7 +46,7 @@
         format: 'json',
         directory: '/content/',
         extension: '.json',
-        collection: ['data','data/people']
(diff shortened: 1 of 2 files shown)


grtjn:
Thnx for catching this! I actually copied that delete from demo-cat, doesn't that work? Not very important, looks good like this!

hunterwilliams:
Delete just seemed weird for accessing the value for the textbox. Since the variable would be gone right? At least normally without angular. 

So I just returned it to its start state.

grtjn:
Angular can be very forgiving.. :P

grtjn:
Worked for me. I'll be using `gulp autotest` from now on!

PR #180, opened 2015-09-11
Fixes #64 - App secret

### LINKED ISSUE #64: Generate new (random) secret key for expressSession
Every app that is generated, starts with expressSession secret key '1234567890QWERTY'. Generate a real random one, or at least make it more unique (by prepending app-name, timestamp or whatever).

hunterwilliams:
Generates a better app secret by using uuid and also now replaces instances of `@sample-app` with `answers.nameDashed` #64

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -1,5 +1,5 @@
 {
-  "name": "ml-node-seed",
+  "name": "@sample-app",
   "version": "0.0.0",
   "dependencies": {
     "angular": "~1.4.4",
@@ -22,5 +22,6 @@
     "angular-mocks": "~1.4.4",
     "sinon": "http://sinonjs.org/releases/sinon-1.16.1.js",
     "bardjs": "~0.1.8"
-  }
+  },
+  "private": true
 }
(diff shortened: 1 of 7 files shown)


grtjn:
Looks very promising, will test today!

grtjn:
Works like a charm!

Can you make a few extra changes?

- `(app/templates/)bower.json` lists name `ml-node-seed`. Can you replace it there as well?
- `(app/templates/)ui/index.html` has an empty `<title>`. Can you insert the app name there too?

Not really related, but would you mind also fixing this while you are at it?

- `(app/templates/)[bower | package | mlpm].json` list different versions. Can you synch them?
- Can you add `private: true` to all three as well?

hunterwilliams:
Would Index.html get appNameDashed or regular app name? Super easy to do either. 

Will knock them all out this weekend or Monday

grtjn:
I had to lookup the diff, but it is just a title, so doesn't really matter too much. I'd be fine either way. It is just to make it a little easier find the right tab back, when you have many open.. :)

hunterwilliams:
All changes should be present

PR #182, opened 2015-09-11
updating the chiclets to show an icon with ban/add

jenbreese:
Updating the selected facets layout.

### DIFF
--- a/app/templates/ui/styles/main.less
+++ b/app/templates/ui/styles/main.less
@@ -12,9 +12,8 @@
 @import "blues-variables.less";
 */
 
-/* MLU design
+/* MLU design */
 @import "mlu.less";
-*/
 
 body {
   background-color: #eee;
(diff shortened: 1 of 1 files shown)


app/templates/ui/styles/main.less
```diff
@@ -12,9 +12,8 @@
 @import "blues-variables.less";
 */
 
-/* MLU design
+/* MLU design */
```
grtjn: Looks like you are enabling MLU design by default. Was that intentional?

app/templates/ui/styles/main.less
```diff
@@ -67,6 +66,33 @@ nav.nav {
...
+}
+
+/* negated and added chiclets */
+.chiclet  {
+    position: relative;
+}
+/* set up the added one first */
+.chiclet:before {
+    content: "\f055";
+    font-family: FontAwesome;
```
grtjn: This is a less file, right? Can't you pull in fa classes here? `.chiclet { .fa; .fa-whatever; }`?

jenbreese: I'm not sure what you mean @grtjn ?

app/templates/ui/styles/main.less
jenbreese: No that is an accident. I will fix.

app/templates/ui/styles/main.less
grtjn: I meant mixins I think. See this example: http://stackoverflow.com/a/1065476/918496

grtjn:
Merging for now to at least have the UI changes applied..

PR #183, opened 2015-09-14
Fixes #145

### LINKED ISSUE #145: Logic of adding email(s) on account page not intuitive
I think it would be better to show an input fields that get added to the list if non-empty. And instead of Add on right, I'd prefer to see a More underneath..

hunterwilliams:
Fixes #145 - I kept the form validation. Let me know if any of this should change. Please test out in the browser. I think it's intuitive now

Also when button said "More" it looked weird so didn't keep that.

### DIFF
--- a/app/templates/node-server/proxy.js
+++ b/app/templates/node-server/proxy.js
@@ -34,8 +34,8 @@ router.put('*', function(req, res) {
   if (req.session.user === undefined) {
     res.send(401, 'Unauthorized');
   } else if (req.path === '/v1/documents' &&
-    req.query.uri.match('/users/') &&
-    req.query.uri.match(new RegExp('/users/[^(' + req.session.user.name + ')]+.json'))) {
+    req.query.uri.match('/api/users/') &&
+    req.query.uri.match(new RegExp('/api/users/[^(' + req.session.user.name + ')]+.json'))) {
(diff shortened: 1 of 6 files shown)


grtjn:
Pity it won't validate changes in existing email addresses, but otherwise a big improvement to what we had. Just need to run a quick test before I'll merge..

grtjn:
Your changes work nicely. They work even better with some extra changes though. This controller stores the profile at /api/users/, but proxy and routes look in the wrong place. Please add `/api` before `/users/` in the following lines:

proxy.js, inside router.put:
```
    req.query.uri.match('/users/') &&
    req.query.uri.match(new RegExp('/users/[^(' + req.session.user.name + ')]+.json'))) {
});
```

routes.js, router.post('/user/login':
```
    path: '/v1/documents?uri=/users/' + username + '.json',
```

I also notice a tiny bug in user.service.js, function updateUser(). The else needs an extra `if ( data.profile.emails ) `, otherwise you get an emails array with a null inside:
```
        if ( _.isArray(data.profile.emails ) ) {
          _currentUser.emails = data.profile.emails;
        }
        else {
          // wrap single value in array, needed for repeater
          _currentUser.emails = [data.profile.emails];
        }
```

hunterwilliams:
Added your changes

grtjn:
Perfect, thnx!

Issue #151, opened 2015-09-03
Single config for ml-host/ml-port


grtjn:
In a format that works for Roxy, gulp, and potentially others in future?

grtjn:
Even better would be something that allows defaults, and env-specific overrides, like Roxy..

grtjn:
Considering patching Roxy using app_specific.rb. Simplest way of reading JSON (using core gems):

```
require 'yaml'
items = YAML.load_file('local.json')
items.each do |key, item|
  puts "#{key}:#{item}"
end
```

And then we can start considering top-level json configs like defaults.json, and local/dev/prod.json, containing something like:

```
{
  "ml-host": "localhost",
  "ml-port": 8040,
  "app-port": 9040
}
```

And potentially a few other things that would be convenient for developers to have readily available..

hunterwilliams:
So I want to discuss this a tiny bit before jumping in.

Currently ports live in:

etc/init.d/node-express-service - Which is a bash script to start it as a service via forever
gulp.config.js - config JSON used by Gulp
deploy/build.properties - used for Roxy

I'm not totally worried about the bash script as it's probably only going to be ran in production or higher envs where a property override wouldn't exist. So that leaves gulp.config.js and build.properties.


I think we need to wait for Roxy to support JSON
* Is Roxy planning to add JSON support? 
  * If so I don't think slush should currently facilitate its work as you won't have overrides working properly without some hacks.
  * If it does or will should it just generate the json for gulp during the slush build? If so, then we just need to add whatever lines it doesn't support. **IDEAL**
* Or should gulp during the server command figure out what build.properties or read from them?
  * This seems like a bad move because we'd be tightly coupled to replicate however roxy does it's properties file decisions. 

Basically I think until Roxy supports building JSON properties this should wait otherwise you'll just get weird behavior. I think changing the ports in 2 places isn't a terrible concern as *most* developers probably will not need to change them as much since now you can do pick different ones during generation.

If it already supports JSON my mistake and I'll implement next week

grtjn:
Anything inside `etc/` is for server deployments only, though you could consider deployment in a local vm a server deployment too. I usually run the node part on the host, and sometimes have ML in a vm. (Depends mostly on which version I need..) So, I agree, and also lean towards ignoring `etc/` in this discussion.

[...]

PR #185, opened 2015-09-15
#151 Single Config

hunterwilliams:
For issue #151

### DIFF
--- a/app/templates/README.mdown
+++ b/app/templates/README.mdown
@@ -42,9 +42,9 @@ On Windows:
 
     ml.bat local deploy content
 
-Edit `gulp.config.js` to set your desired ports
+Edit `./gulp-local.json` to set your desired ports
 
-    gulp serve-dev # this will watch the .less file for changes, compile them to .css, and run the node server
+    gulp serve-local # this will watch the .less file for changes, compile them to .css, and run the node server
 
 # Installing as service
 
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -6,6 +6,7 @@ var args = require('yargs').argv;
 var browserSync = require('browser-sync');
 var config = require('./gulp.config')();
 var del = require('del');
+var fs = require('fs');
 var glob = require('glob');
 var gulp = require('gulp');
 var path = require('path');
@@ -160,6 +161,31 @@ gulp.task('inject', ['wiredep', 'styles', 'templatecache'], function() {
     .src(config.index)
     .pipe(inject(config.css))
     .pipe(gulp.dest(config.client));
+});
+
+
+/**
+ * Creates a sample gulp-local.json; can be used as model for other environments
+ */
+gulp.task('init-local', function() {
+  //copy from slushfile - config gulp - with modifications to use config instead
+  log('Creating gulp-local.json sample document with values drawn from gulp.config.js');
+  try {
+    var configJSON = {};
+    configJSON['ml-version'] = config.marklogic.version;
+    configJSON['ml-host'] = config.marklogic.host;
+    configJSON['ml-http-port'] = config.marklogic.httpPort;
+    configJSON['node-port'] = config.defaultPort;
+
+    if (config.marklogic.version < 8) {
+      configJSON['ml-xcc-port'] = config.marklogic.xccPort;
+    }
+
+    var configString = JSON.stringify(configJSON, null, 2) + '\n';
+    fs.writeFileSync('gulp-local.json', configString, { encoding: 'utf8' });
+  } catch (e) {
+    console.log('failed to write gulp-local.json: ' + e.message);
+  }
 });
 
 /**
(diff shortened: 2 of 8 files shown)


app/templates/README.mdown
```diff
@@ -42,9 +42,9 @@ On Windows:
 
     ml.bat local deploy content
 
-Edit `gulp.config.js` to set your desired ports
+Edit `./deploy/gulp-local.json` to set your desired ports
```
grtjn: you can put the gulp-local.json in the root, instead of under deploy/

app/templates/gulpfile.js
```diff
@@ -473,17 +482,27 @@ function serve(isDev, specRunner) {
...
+ * Determine whether the enviornment requires dev mode'
+ * @param {String} env - 'local', 'dev', or 'prod'
+ * @return {Boolean} - true if env==='local'
+ */
+function isDevMode(env) {
+  return env === 'local';
+}
+
+function getNodeOptions(env) {
+  var envJson = require('./deploy/gulp-' + env + '.json');
```
grtjn: Won't the require break if gulp-local.json doesn't exist? If yes, perhaps try to ignore the error, and maybe print a message suggesting you can provide local overrides with that file.

slushfile.js
```diff
@@ -260,34 +273,14 @@ gulp.task('generateSecret', ['init'], function(done) {
 gulp.task('configGulp', ['init'], function(done) {
 
   try {
+    var configJSON = {};
+    configJSON['ml-host'] = 'localhost';
```
grtjn: Can you make the slush generator ask for the host as well?

hunterwilliams:
@grtjn I've added the change so it doesn't need the xccport for ml8.

also I renamed ml-port to ml-http-port because it's a good clarification I think for inexperienced users

app/templates/gulpfile.js
```diff
@@ -439,12 +448,12 @@ function orderSrc (src, order) {
  * serve the code
  * --debug-brk or --debug
  * --nosync
- * @param  {Boolean} isDev - dev or build mode
+ * @param  {String} env - 'dev','test', or 'prod'
```
grtjn: Small typo? env = local | dev | prod..

slushfile.js
```diff
@@ -260,34 +289,19 @@ gulp.task('generateSecret', ['init'], function(done) {
...
+    if (settings.mlVersion < 8) {
+      configJSON['ml-xcc-port'] = settings.xccPort;
+    }
 
-    fs.writeFileSync('etc/init.d/node-express-service', nodeService);
+    var configString = JSON.stringify(configJSON, null, 2) + '\n';
+    fs.writeFileSync('gulp-local.json', configString, { encoding: 'utf8' });
   } catch (e) {
-    console.log('failed to update node-express-service: ' + e.message);
+    console.log('failed to write gulp.local.json: ' + e.message);
```
grtjn: Tiny typo, gulp-local.json..

app/templates/gulpfile.js
```diff
@@ -473,17 +482,35 @@ function serve(isDev, specRunner) {
...
+
+function getNodeOptions(env) {
+  var envJson;
+  var envFile = 'gulp-' + env + '.json';
+    try {
+      envJson = require(envFile);
+    }
+    catch (e) {
+      envJson = {};
+      console.log('Info: ' + envFile + ' - may not exist. Did you mean to create one?');
```
grtjn: A different project was using a similar pattern for a `local.gradle` file. If missing it printed:

"Couldn't find local.gradle; you can create this gitignore'd file to override properties if necessary"

Maybe print a similar message in our case?

grtjn: A thought just struck me. People will ask themselves, how should gulp-local.json look like? How about adding a gulp init-local task that helps create that file, and refer to that in the info message?

slushfile.js
```diff
@@ -260,34 +289,19 @@ gulp.task('generateSecret', ['init'], function(done) {
 gulp.task('configGulp', ['init'], function(done) {
 
   try {
+    var configJSON = {};
+    configJSON['ml-host'] = settings.marklogicHost;
```
grtjn: tiny favor: can you add ml-version in here too?

grtjn:
Good work, really looking forward to this.. :)

hunterwilliams:
This does not cover proxy.js and routes.js properly for port changing looking into that now

hunterwilliams:
Fixed that. The change to of env.ML_PORT to env.ML_HTTP_PORT broke it as I wasn't populating it in getNodeOptions(). I changed it back since in getNodeOptions() it looks for env.ML_PORT and I'm not sure where that value is supposed to come from

Thoughts?

grtjn:
env.ML_PORT is used between gulpfile.js and node-app.js mostly. Looks good enough as is.

There seems to be a merge conflict, can you rebase? I'll have a spin with these changes after rebase..

grtjn:
Yay, didn't need to change generated local.properties and gulp-local.json to run ml local bootstrap and gulp serve! Really happy with this..

Just need to remind myself to not run gulp serve-dev, but `gulp serve-local` from now on!

PR #189, opened 2015-09-16
Fixed #31: moved bodyparsing to allow upload of non-json

### LINKED ISSUE #31: Issue with proxying data to ML

grtjn:
#31

### DIFF
--- a/app/templates/node-server/node-app.js
+++ b/app/templates/node-server/node-app.js
@@ -4,7 +4,6 @@
 var express = require('express');
 var expressSession = require('express-session');
 var app = express();
-var bodyParser = require('body-parser');
 var logger = require('morgan');
 var port = process.env.PORT || 8001;
 var four0four = require('./utils/404')();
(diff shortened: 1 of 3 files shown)


grtjn:
Needs thorough testing. Saw some weird issues with ml-highcharts. See also #177..

grtjn:
I think I need to take a look at this commit from @ryanjdew, he moved the bodyparser inside router. Interesting.. https://github.com/ryanjdew/slush-marklogic-node/commit/962554609ad168bf12b435c355955cea9be31373

grtjn:
This looks better, but still needs more testing..

grtjn:
Missed out on a few important changes to proxy.js, added those. Even a simple search call from UI to backend failed, with these extra changes all seemed fine..

grtjn:
Works fine with JSON updates, so at least no regression. Would be nice to test this with binary upload too, though..

grtjn:
Noticed another glitch with the out of the box app-user. It gets a 403, which wasn't handled properly in /user/login. Fixed that.

grtjn:
Proper app-user privs for accessing UI can be found in #298. Comes down to rest-reader/rest-writer priv..

grtjn:
Tested this sufficiently, going to merge this PR, as it contains some important fixes..

PR #192, opened 2015-09-17
#191 Users in local.json fix

### LINKED ISSUE #191: Overriding users in local.json disregarded
As title says. Will cause issues in node if changed in local.json and not gulp config.

hunterwilliams:
Fixes #191 
However current issue of not suing the appuser as expected

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -523,14 +523,14 @@ function isDevMode(env) {
 
 function getNodeOptions(env) {
   var envJson;
-  var envFile = './gulp-' + env + '.json';
+  var envFile = './' + env + '.json';
     try {
       envJson = require(envFile);
     }
     catch (e) {
       envJson = {};
       console.log('Couldn\'t find ' + envFile + '; you can create this file to override properties - ' +
-        '`gulp init-local` creates gulp-local.json which can be modified for other environments as well');
+        '`gulp init-local` creates local.json which can be modified for other environments as well');
     }
   var port = args['app-port'] || process.env.PORT || envJson['node-port'] || config.defaultPort;
   return {
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -165,15 +165,19 @@ gulp.task('inject', ['wiredep', 'styles', 'templatecache'], function() {
 
 
 /**
- * Creates a sample gulp-local.json; can be used as model for other environments
+ * Creates a sample local.json; can be used as model for dev.json and prod.json
  */
 gulp.task('init-local', function() {
   //copy from slushfile - config gulp - with modifications to use config instead
-  log('Creating gulp-local.json sample document with values drawn from gulp.config.js');
+  log('Creating local.json sample document with values drawn from gulp.config.js');
   try {
     var configJSON = {};
     configJSON['ml-version'] = config.marklogic.version;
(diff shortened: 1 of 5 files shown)


app/templates/gulpfile.js
```diff
@@ -165,15 +165,19 @@ gulp.task('inject', ['wiredep', 'styles', 'templatecache'], function() {
...
   //copy from slushfile - config gulp - with modifications to use config instead
-  log('Creating gulp-local.json sample document with values drawn from gulp.config.js');
+  log('Creating local.json sample document with values drawn from gulp.config.js');
   try {
     var configJSON = {};
     configJSON['ml-version'] = config.marklogic.version;
     configJSON['ml-host'] = config.marklogic.host;
+    configJSON['ml-admin-user'] = config.marklogic.username;
+    configJSON['ml-admin-pass'] = config.marklogic.password;
+    configJSON['ml-app-user'] = config.marklogic.username;
```
hunterwilliams: This needs to pull from build.properties but apparently the app user doesn't current have rest permissions so therefore can't run slush..

grtjn:
Looks very interesting. The change from gulp-local.json to local.json makes sense as the json is meant to be used by Roxy as well (hopefully soon). The admin/admin is now defined in two places: roxy properties, and gulp.config. Proxy.js is using that as well, but it would be better if that would switch to app-user instead. Unfortunately, the app-user currently doesn't get the privs it needs to actually run the app. Hoping to look into that as well.

I'll be opening Roxy tickets for that asap..

app/templates/gulpfile.js
grtjn: I think we could/should go the other way around. Name of default-user is predictable, so we can deduce that (or we could just replace it), and we could generate a new password, and override the one generated by Roxy..

hunterwilliams: Either way. I can replace it here. Just mostly don't want to fix it here and then end up preemptively using the appuser which doesn't have proper roles yet

grtjn: Yes, too early just now, but that is how we can do it when we have fixed the app-user privs..

slushfile.js
```diff
@@ -324,6 +333,8 @@ gulp.task('init', ['checkForUpdates'], function (done) {
   var prompts = [
     {type: 'list', name: 'mlVersion', message: 'MarkLogic version?', choices: ['8','7', '6', '5'], default: 0},
     {type: 'input', name: 'marklogicHost', message: 'MarkLogic Host?', default: 'localhost'},
+    {type: 'input', name: 'marklogicAdminUser', message: 'MarkLogic Admin User?', default: 'admin'},
+    {type: 'input', name: 'marklogicAdminPass', message: 'MarkLogic Admin Password?', default: 'admin'},
```
grtjn: Can you make the default for the admin password blank? Something you can't enter yourself, and works with current serve-local too. It is configured such that you need to call /api/user/login first to initialize session, so default user is not used yet. (It only becomes relevant if you want to disable auth, like I did for github-search..)

grtjn:
Found one omission. This line still mentions `gulp-`: https://github.com/hunterwilliams/slush-marklogic-node/blob/user-fix/app/templates/gulpfile.js#L526. Otherwise this PR works fine..

grtjn:
Good to go, any comments?

grtjn:
No comments so far, merging..

PR #203, opened 2015-09-18
#194 Binary corruption fix in gulp replace

### LINKED ISSUE #194: Binary Corruption
During a slush generation, it appears that the images or zips can be corrupted.

Windows 10
Latest Master Of Slush

Images are not corrupted in the generator locally but get corrupted during slush process

hunterwilliams:
Fixes #194 

Quick google search found: https://github.com/lazd/gulp-replace/issues/6

So they added a binary option. I added the use of it here. Unfortunately not a good way to catch this error except by hand. Good eyes @jenbreese

### DIFF
--- a/slushfile.js
+++ b/slushfile.js
@@ -382,7 +382,7 @@ gulp.task('init', ['checkForUpdates'], function (done) {
             }
 
           }))
-          .pipe(replace('@sample-app',answers.nameDashed))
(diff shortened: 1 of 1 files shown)


jenbreese:
This is great. Thanks for figuring it out @hunterwilliams!

grtjn:
Works, merging!

PR #217, opened 2015-09-23
Separated out the options into their own util

hunterwilliams:
Think this will become an issue soon. Ran into it trying to move stuff around for passport integration


hunterwilliams:
Removing. Will occur as part of #220

Issue #175, opened 2015-09-09
Use Passport For Login


hunterwilliams:
As a developer, I would like to be able to swap out logins with more common ones vs necessarily using MarkLogic users.

joemfb:
I've got an implementation using passport here: https://github.com/joemfb/mlpm-registry/blob/master/server.js. It's poorly factored and messy code, but could be a useful starting point.

grtjn:
Assigning this to milestone 0.2.1, nice to have..

grtjn:
Related to #284, and incorporated in PR #285 which supersedes PR #220..

grtjn:
Fixed in master..

PR #220, opened 2015-09-23
#175  Passport

hunterwilliams:
#175 Using as a placeholder. Also so anyone can take a peek if they have any tips. I have more of the development locally... but avoiding committing because a lot I think I'll have to back out

@joemfb - know you were interested

### DIFF
--- a/app/templates/node-server/routes.js
+++ b/app/templates/node-server/routes.js
@@ -5,29 +5,28 @@
 var router = require('express').Router();
 var four0four = require('./utils/404')();
 var http = require('http');
-var config = require('../gulp.config')();
+var options = require('./utils/options')();
+var auth = require('./utils/auth');
 
-var options = {
-  appPort: process.env.APP_PORT || config.defaultPort,
-  mlHost: process.env.ML_HOST || config.marklogic.host,
-  mlHttpPort: process.env.ML_PORT || config.marklogic.httpPort,
-  defaultUser: process.env.ML_APP_USER || config.marklogic.user,
-  defaultPass: process.env.ML_APP_PASS || config.marklogic.password
-};
+auth.init();
+
+router.post('/user/login', auth.handleLocalAuth);
 
 router.get('/user/status', function(req, res) {
   var headers = req.headers;
   noCache(res);
-  if (req.session.user === undefined) {
+
+  if (!req.isAuthenticated()) {
     res.send({authenticated: false});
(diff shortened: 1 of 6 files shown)


hunterwilliams:
Serverside is done and works as previously except possible weirdness with updating the user, but it could have been odd before.

app/templates/node-server/routes.js
```diff
@@ -4,100 +4,43 @@
...
+    user = req.session.passport.user;
+  }
+  else if (req.session.user)
+  {
+    user = req.session.user;
+  }
+  if (!user) {
     res.send('{"authenticated": false}');
   } else {
+    var name = user.username || user.name;
```
hunterwilliams: this whole method may just need some cleanup

hunterwilliams:
Looks like I can swap out the redirect and basically make the login service back to what it was. Just need to implement this -> http://toon.io/on-passportjs-specific-use-cases/ and use a different ``passport.authenticate`` method to return json like it used to.

I'll do this later. If you have any input @joemfb on that or any other thoughts before I do this then prettify it - mostly pertaining to allow it to support other logins easily.

hunterwilliams:
Managed to get it working as said without changes to ui. However, it'd be significantly harder to drag and drop in a new login strategy. So going to try to make the ui adapt to the server changes instead.

app/templates/node-server/utils/auth.js
```diff
@@ -0,0 +1,102 @@
...
+            if (json.user !== undefined) {
+              user.profile = json.user;
+            } else {
+              console.log('did not find chunk.user');
+            }
+
+            done(null, user);
+          });
+        } else if (response.statusCode === 401) {
+          done(null, false, {message: 'Invalid credentials'});
```
hunterwilliams: Actually may need to change this to saying logged in just has no user profile..

hunterwilliams:
Fixed a bug with it not allowing logins without existing profiles. Got it working with the current master. So it's ready to be merged.

grtjn:
Looks like there is a lot of good stuff in this PR. Anyone more familiar with Passport and such than me that likes to take a look?

grtjn:
These Passport changes were (manually) merged into PR #285, closing this one. Thanks Hunter for your work!

hunterwilliams:
Haha that's awesome! Can't imagine it was a fun merge

PR #230, opened 2015-09-24
Fix #210 HighlightJS error

### LINKED ISSUE #210: Unable to display xml data when using hljs in detail.html
I'm using some XML data and want to display the content when selected from a search. In previous iterations of slush-marklogic-node I made the following change to detail.html:

    <div class="row detail">
      
      <div hljs source="model.detail" language="xml"></div>
    </div>

and added hljs to app.js, but this does not work with the latest version of slush-marklogic-node I am using (0.2.0), with no output being displayed. The code I used is:

    <div class="panel-body">
        <div hljs source="ctrl.doc" language="xml"></div>
    </div>

but I did manage to get this working:

    <div class="panel-body">
        <pre>{{ctrl.doc}}</pre>
    </div>

hunterwilliams:
#210 

Moved highlightjs above angular-highlightjs to hopefully ensure get correct order. I don't like this solution since it's easy to screw it back up. 

Could possibly not autoinject highlights.js script and move it outside of the normal js includes. Not sure what's ideal here..

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -10,13 +10,17 @@
     "angular-ui-tinymce": "~0.0.9",
     "bootstrap": "~3.3.5",
     "font-awesome": "~4.4.0",
-    "highlightjs": "~8.7.0",
     "jquery": "~2.1.4",
     "lodash": "~3.10.1",
     "ml-search-ng": "~0.2.0",
     "ml-utils": "withjam/ml-utils",
     "ng-json-explorer": "8c2a0f9104",
     "angular-mocks": "~1.4.4"
+  },
+  "overrides": {
+    "angular-highlightjs": {
+      "dependencies": {"angular" : ">1.0.8", "highlightjs":"~8.7.0"}
+    }
   },
   "devDependencies": {
     "angular-mocks": "~1.4.4",


hunterwilliams:
May be smarter to do an oddball override -> https://github.com/taptapship/wiredep#bower-overrides

grtjn:
Yeah, I noticed the importance of the order with other libraries as well. Maybe we should as a general pattern put all angular-* dependencies at the bottom.

Not sure how bower overrides would help here, unless by defining proper dependencies (if lacking in angular-highlightjs)..

hunterwilliams:
correct. could override the dependencies

hunterwilliams:
Unfortunately overriding means getting rid of existing dependencies which seems like a really bad move... Not sure the best option here. 

@grtjn my only worry with moving all of the dependencies to the bottom is someone will go along later and be like why aren't these alphabetized - eh let's do it and then stuff starts breaking without notice

grtjn:
I don't think many will alphabetize the list, I was the first to . ;), and if we have it semi alphabetized, with angular visibly at the bottom, hopefully people will think that was for a reason.. :P

But I do agree, both approaches are a little fragile. Best would be to file a ticket against angular-highlightjs.

I'd vote doing all though, change order, bower override, and ticket. The bower override only as a temp solution, with a ticket to remind ourselves to get rid of it at some point..

hunterwilliams:
I was going to add a ticket to them but the developer stated that haven't picked due to lack of official one. 

So a bower override with all dependencies? Can possibly make the angular-highlights-js and highlights-js only point to the very current version

hunterwilliams:
Done. I don't want to muck with the order of dependencies anymore in case I break something unintentionally like this

grtjn:
Works for me!

Issue #235, opened 2015-09-24
Clear search invisible with Blue design


grtjn:
@jenbreese Can you take a look at the search bar with the blue design enabled? I don't see the (x) icon to click after entering a search term, though it looks like I can click something in the blind. The cursor changes, and the search bar does get cleared..

Not super important, just happened to notice..

PR #237, opened 2015-09-24
fix for #235

jenbreese:
made the circle with the x visible for bug #235


jenbreese:
I added a smidge more spacing to the clear. It looked tighter because the blues design.

grtjn:
Works for me!

PR #238, opened 2015-09-25
Fix #229 Account Edits

### LINKED ISSUE #229: Account Edits Not Properly Saved
Affects Master

Not a very big deal as it doesn't affect functionality negatively. Have to get in weird use case

Recreation
1. Login
2. Edit Account details and submit
3. Verify new details on page
4. Refresh the page & now changes appear to be *missing*
5. Logout & log back in
6. Changes are now *present*

Behind the scenes
* Edits are saved locally and then sent to the server
* Since edits are local they are dropped on refresh but live in the database
* /api/user/status serves the user from session variable and is only updated upon login

Potential issue if user edits account refreshes -> then will not see edits or will override existing edits if they hit submit again.

hunterwilliams:
Basically changed /api/user/status to work properly
Fixes #229

### DIFF
--- a/app/templates/node-server/routes.js
+++ b/app/templates/node-server/routes.js
@@ -16,14 +16,44 @@ var options = {
 };
 
 router.get('/user/status', function(req, res) {
+  var headers = req.headers;
   noCache(res);
   if (req.session.user === undefined) {
-    res.send('{"authenticated": false}');
+    res.send({authenticated: false});
   } else {
-    res.send({
-      authenticated: true,
-      username: req.session.user.name,
-      profile: req.session.user.profile
+    delete headers['content-length'];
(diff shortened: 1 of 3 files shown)


app/templates/node-server/routes.js
```diff
@@ -16,14 +16,44 @@ var options = {
...
+          if (json.user !== undefined) {
+            var profile = {
+              fullname: json.user.fullname,
+              emails: json.user.emails
+            };
+            res.status(200).send({
+              authenticated: true,
+              username: req.session.user.name,
+              profile: profile
+            });
```
grtjn: /user/login stores the profile in the session. Can you synch with that?

grtjn:
Works for me!

PR #249, opened 2015-09-28
Fixes #248 detail controller test breaking build

### LINKED ISSUE #248: Unable To Build Out of Box/Detail Controller Test Fails
Affects: Master/1.0

gulp build fails due to detail.controller.spec.js containing a test that was never updated since the changes of detail.controller.js. The only way to get a successful build is to comment out the test.

hunterwilliams:
Fixes #248 

Partially mocked the document object to make it follow the controller methods it would hit


grtjn:
A new issue with login.service.js breaks gulp build tests, but apart from that this fix works just fine. Merging this..

PR #266, opened 2015-11-11
Add dynamic date/dateTime selectors

ryanjdew:
Tied to issue #265

### DIFF
--- a/app/templates/ui/app/search/search.controller.js
+++ b/app/templates/ui/app/search/search.controller.js
@@ -27,5 +27,86 @@
     $scope.$watch(userService.currentUser, function(newValue) {
       ctrl.currentUser = newValue;
     });
+
+    /* BEGIN Date/DateTime constraint logic */
+    ctrl.dateFilters = {};
+    ctrl.dateStartOpened = {};
+    ctrl.dateEndOpened = {};
+    ctrl.pickerDateStart = {};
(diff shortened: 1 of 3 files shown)


grtjn:
Very interesting stuff! Maybe worth pushing into ml-search itself? That would also allow adding specific support for those dateFilters inside ml-search, and prevent using additionalQueries.

ryanjdew:
That sounds good to me. @joemfb, what do you think? Do you think it makes sense to move into ml-search?

joemfb:
I think it'd be good to have the functionality in ml-search-ng, but I'm not sure exactly where it should go. I guess it wouldn't be too hard to bake it into the search controller prototype. Longer term, I'd like to change the way facets/constraints are handled in the search context, to better support these kinds of features.

Do you have a (rough) plan for how you'd integrate this feature?

grtjn:
I think you'd need:

- Support for operators on facet selections. The default would be EQ of course, but there should be an option to pass in NE, LT, LE, GT or GE too.
- Awareness of facet type perhaps, so that you can send in the selected value in the appropriate data format.
- An ml-facets template that considers the facet type, and responds to type date/dateTime by showing start/end pickers, connected with the right toggle/selectFacet callback to get the correct facet selection added to the list of filters..

Issue #245, opened 2015-09-25
XML preview in detail page not using syntax highlights


grtjn:
Copied from #131:

Hmm, PR #230 makes it possible to use hljs instead of pre. If you enable the 'hljs' library in detail.module.js, and then add `hljs source="ctrl.xml" language="xml"` to the `<pre>` elements, you can see how that looks. Unfortunately hljs like putting things in backgrounds, which is what the pre-xml class is trying to get rid off.

@jenbreese WDYT, does it make sense to use hljs, and can we fix the presentation? Or stick with the pre for the moment? hljs just adds color, nothing else..

In more general (to all), also thinking, is there a better library than hljs for viewing xml? With not only color, but also collapsable like the json-explorer is showing json?

jenbreese:
@grtjn Let's talk about this tomorrow?

grtjn:
You can ping me on IM.. :)

janmichaelyu:
PR #276 should resolve this. I discovered that using
```html 
<pre source="ctrl.xml" language="xml"></pre>
```
wasn't working and some hljs documentation was outdated.

grtjn:
I notice now that my description in the top didn't include the `hljs` in the suggested addition to `<pre>`. That attribute is essential..

grtjn:
One year further, and a lot has happened in between. I have recently released the view-file-ng component (https://grtjn.github.io/view-file-ng/#/), would that be a better solution for out of the box display of json, xml, text, and binary, both raw and pretty, and with a download button as well?

@janmichaelyu / @jenbreese WDYT?

grtjn:
Discussed with @jenbreese offline, we are adding view-file-ng in cards theme release..

grtjn:
Fixed in cards-theme branch

PR #276, opened 2015-12-14
245 - fix xml highlighting using hljs and github.css as default stylesheet

janmichaelyu:
#245 Sample display for native xml document:

2015-12-14 18_02_16-slush-test

Sample display for converted json document
2015-12-14 18_02_27-slush-test

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -28,6 +28,12 @@
   "overrides": {
     "angular-highlightjs": {
       "dependencies": {"angular" : ">1.0.8", "highlightjs":"~8.7.0"}
+    },
+    "highlightjs": {
+      "main": [
+        "highlight.pack.js",
+        "styles/github.css"
+      ]
     }
   },
   "devDependencies": {
--- a/app/templates/ui/app/detail/detail.controller.spec.js
+++ b/app/templates/ui/app/detail/detail.controller.spec.js
@@ -1,8 +1,8 @@
 /* jshint -W117, -W030 */
-(function () {
+(function() {
   'use strict';
 
-  describe('Controller: DetailCtrl', function () {
+  describe('Controller: DetailCtrl', function() {
 
     var controller;
     var doc;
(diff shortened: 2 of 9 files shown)


app/templates/bower.json
```diff
@@ -12,7 +12,6 @@
     "angular-x2js": "https://github.com/janmichaelyu/angular-x2js.git",
     "bootstrap": "~3.3.5",
     "font-awesome": "~4.4.0",
-    "highlightjs":"~8.7.0",
```
grtjn: This looks a bit suspicious, why delete this dependency? I think it is necessary. Run `bower prune` to make sure you are not running against dependencies that are no longer in bower.json..

janmichaelyu: Thanks for spotting this. I've put the dependency back in.

janmichaelyu:
@grtjn Please check and merge.

grtjn:
Looks pretty clean to me. Wondering though if there is an easier/cleaner way to get rid of the well..

@jenbreese What do you think about the looks of XML like this? (screenshots above)

jenbreese:
@grtjn I think it looks clean. I'm not sure what you mean by the word "well"?

grtjn:
'well': http://www.w3schools.com/bootstrap/bootstrap_wells.asp, the background that highlightjs puts behind the pretty-printed code: http://pc035860.github.io/angular-highlightjs/example/#/hljs. Mike uses 15 lines of less/css code to undo the default pre-styling. Is there a simpler/shorter way of getting the same? Note: the wrapping element doesn't have to be a `<pre>`..

grtjn:
Can you rebase, and verify these changes will apply to all ui themes?

janmichaelyu:
Rebased and updated. Please check.

grtjn:
@jenbreese I think it makes most sense if you review this work. Can you generate an app with this, and see how it looks?

I was also wondering if there is a way to simplify the css changes. But maybe there isn't?

grtjn:
Pretty much obsolete due to integration of view-file-ng. I also copied over some pieces in other PR's, like test changes for detail.

Using github highlightjs theme instead of the default one might be a nice touch though, even though it seems a subtle diff. @jenbreese WDYT? See https://highlightjs.org/static/demo/

grtjn:
Mike, closing this PR. We have taken elements out of this, and processed them separately. Sorry for your trouble, but thanks for the effort. There were definitely useful elements in this PR..

PR #294, opened 2016-01-29
Fixed #269: fixed init script, and added docs/scripts for server deployment

### LINKED ISSUE #269: included init scripts and forever.js not working
Specifically stop/restart doesn't actually kill the app, or remove the PID file. I've needed to manually kill the app and forever.js as well as remove the PID file to restart. I'll add more details when I can.

grtjn:
#269


grtjn:
This includes a bugfix, and some very convenient scripts. Merging now..

Issue #246, opened 2015-09-25
Prepend app-specific ML users with app-name


grtjn:
And make middle-ware add that automatically to user names, so end users don't need to type the prefix in the login box..

It can be seen as a good practice to prevent user conflicts when deploying multiple demos on a single server.

hunterwilliams:
So they'd just type user?  And it'd make it sample-app-user?  That seems quite odd behavior for tracking issues later

grtjn:
It prevents all kinds of potential conflicts with users between multiple demos (deployed in one server), as well as conflicts with built-in users like 'nobody', 'admin', etc. How do you mean odd?

hunterwilliams:
I guess we prevent them from having to type sample-app-user, but from an ops perspective they say they're using one name you're verifying another on MarkLogic. Also if you need another user for an app you have to remember to name it correctly or it won't work. Room for human or middleware error to save a couple keyboard presses once before they save their password to the browser.

grtjn:
It is not really meant for saving typing characters in the UI, but prevent conflicts with users defined in ML for other purposes. Now if people create custom/adhoc users for a demo they tend to not add such a prefix. Adding such a prefix is just a good practice. It does mean that the dev or operator needs to keep in mind that such a prefix is necessary, but that is just a matter of education.

Note: samplestack does it as well..

grtjn:
PR #298 allows enabling a flag in the middle-tier that forces login to only accept user names that start with app-name + '-'. It is optional though, and not enabled by default.

PR #298, opened 2016-01-30
Fixed #288: allow guest-acces, disallow updates, require app users only

### LINKED ISSUE #288: Make it easier to disable login/auth
Projects like ml-github-search don't need login at all, but disable login/auth is a bit of a hassle at the moment. I discovered the shortest way out is to make the middle-tier always use defaultUser, and force a session.user (optionally with dummy profile). You can very easily do this with a small function that is used as custom middleware in the app.use chain. It could be commented by default, and one would only merely need to uncomment it to disable login..

grtjn:
#288 

Uncomment few lines in node-app.js, and you by-pass auth entirely. Streamlined default app-user stuff to make it work out of the box as Guest with rest-reader/writer privs..


slushfile.js
```diff
@@ -275,8 +292,8 @@ gulp.task('configGulp', ['init'], function(done) {
     configJSON['ml-host'] = settings.marklogicHost;
     configJSON['ml-admin-user'] = settings.marklogicAdminUser;
     configJSON['ml-admin-pass'] = settings.marklogicAdminPass;
-    configJSON['ml-app-user'] = settings.marklogicAdminUser; //THIS NEEDS TO CHANGE
-    configJSON['ml-app-pass'] = settings.marklogicAdminPass; //THIS NEEDS TO CHANGE
+    configJSON['ml-app-user'] = settings.appName + '-user';
```
joemfb: Could this be read from Roxy properties instead of concatenated?

grtjn: Probably, but sounds like a lot of complexity. And still not convinced Node should depend on Roxy configs. I'd prefer the other way round..

Keep in mind this piece of code is only executed at start of project, directly after executing Roxy new, so naming of default-user is reliable at that point. If you would want to make it fully dynamic, gulp serve and service scripts would have to read Roxy props too..

grtjn:
I ran an elaborate test myself, but could use a test run from someone else to confirm. Had to streamline quite a bit..

slushfile.js
```diff
@@ -233,6 +238,18 @@ function configRoxy() {
...
+      '<privileges>\n' +
+      '        <privilege>\n' +
+      '          <privilege-name>rest-reader</privilege-name>\n' +
+      '        </privilege>\n' +
+      '        <!-- remove the rest-writer privilege if read-only access is required -->\n' +
+      '        <privilege>\n' +
+      '          <privilege-name>rest-writer</privilege-name>\n' +
+      '        </privilege>\n' +
+      '      </privileges>\n');
+
```
grtjn: Using rest-reader/writer privs also requires making sure REST extensions and such get app-role permissions, or otherwise calls those REST extensions will fail. This means patching app_specific with something like:

```
  alias_method :original_deploy_rest, :deploy_rest
  def deploy_rest
    original_deploy_rest

    r = execute_query(%Q{
        xquery version "1.0-ml";

        for $uri in cts:uris()
        return (
            $uri,
            xdmp:document-set-permissions($uri, (
                xdmp:permission("#{@properties["ml.app-name"]}-role", "read"),
                xdmp:permission("#{@properties["ml.app-name"]}-role", "execute")
            ))
       )
      },
      { :db_name => @properties["ml.modules-db"] }
    )
    r.body = parse_json r.body
    logger.debug r.body
  end
```

slushfile.js
grtjn: Safe for slush generator, as the line is commented out in build.properties by default, but worth considering for the (new) gulp init-local/dev/prod..

slushfile.js
grtjn: Decided to include a pre-made app_specific.rb in templates to solve this. Showcases some other nice gimmicks as well..

slushfile.js
grtjn: Leveraging the relatively new `ml env info --format=json` in gulpfile..

grtjn:
Ran elaborate tests with this, merging now..

grtjn:
This also fixed #415..

grtjn:
This also fixes #246..

PR #300, opened 2016-01-31
Fixed #295: removed cookie-parser in favor of newer express-session

### LINKED ISSUE #295: Drop cookieParser
cookieParser was necessary for expressSession in the past, but that has long since added own support for cookies. It also seems to have been used in wrong order anyhow. I vote we drop cookieParser, and raise the min dependency for express-session to ^1.5.0. See also:

http://stackoverflow.com/questions/27961320/when-should-i-use-cookie-parser-with-express-session

grtjn:
#295 

I thought I tested this, but as part of something else. Won't harm to test explicitly..

### DIFF
--- a/app/templates/node-server/node-app.js
+++ b/app/templates/node-server/node-app.js
@@ -5,7 +5,6 @@ var express = require('express');
 var expressSession = require('express-session');
 var app = express();
 var bodyParser = require('body-parser');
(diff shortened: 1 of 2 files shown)


grtjn:
Still needs more testing..

grtjn:
Tested in an existing demo, works fine..

PR #304, opened 2016-02-10
Fix uri resolution on the detail page

### LINKED ISSUE #303: detail router stripping '/' from uri on page refresh
I confirmed this with a fresh application generated with bleeding-edge slush. When I load the sample data, eg, the page loads fine. 

Though, for reasons I don't understand, it calls out to the REST API for the document twice. The first time is successful, but the second time, the URI is truncated: the preceding '/' is dropped.

The successful first call looks like:

    http://localhost:3000/v1/documents?format=json&uri=%2Fdata-922.json

The second call drops the slash and so returns a 404:

    http://localhost:3000/v1/documents?format=json&uri=data-922.json

If I refresh the page, or deep-link to it, I always get the second, broken version, and the page fails to load.

I confirmed that the [uri in the `resolve` function in the angular router's detail state definition](https://github.com/marklogic/slush-marklogic-node/blob/master/app/templates/ui/app/route/routes.js#L85) is incorrect (missing the 'slash') in these cases. The culprit seems to be [`squash: true`](https://github.com/marklogic/slush-marklogic-node/blob/master/app/templates/ui/app/route/routes.js#L76) ... everything works if I remove that. The odd thing is that this line has been [present since July](https://github.com/marklogic/slush-marklogic-node/commit/a36cb1dd62ffa384bf9975b97ba5c935152441e1#diff-915a6ab24ca76072e0af4a76992f33d0R86).

I am going to submit a pull request to delete that line. It seems to be working fine without it. According to the [docs on squash](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$stateProvider), it is intended to clean up URLs. Perhaps in a older version of the router library, it helped clean up double-slashes (something like /detail//docs/1234.xml), but it doesn't seem necessary any longer.

patrickmcelwee:
Fixes #303

### DIFF
--- a/app/templates/ui/app/route/routes.js
+++ b/app/templates/ui/app/route/routes.js
@@ -73,7 +73,6 @@
         url: '/detail{uri:path}',
         params: {
           uri: {
-            squash: true,
             value: null
           }
         },


grtjn:
Needs a second test to confirm, though looks good..

grtjn:
Works good for me (tested in existing and new project)..

PR #306, opened 2016-02-15
Fixed #206: add ms visual studio flag for compiling

### LINKED ISSUE #206: Npm install of browser-sync seems to break on Windows (7)
Browser-sync depends on components like node-gyph, that need Visual C++ runtime libraries: http://www.browsersync.io/docs/#windows-users

Looks like you need to make sure to have that available. Seen at least one report where running latest slush broke at browser-sync on Windows 7. Maybe because of this, not entirely sure..

Anyone else seen this? I'll see if I can run a test on a win8 machine..

grtjn:
#206.

Need to run a quick test to see if npm install doesn't break on non-win. And pref someone on a clean win to check if it works..


grtjn:
Runs smooth for me on my mac (so no side-effects for other OS).

Anyone care to run a test on Windows?

grtjn:
(forgot to push a change before writing the comment)

grtjn:
Anyone around with a Windows machine that likes to test this?

PR #313, opened 2016-02-19
Fixed #262: making sure stdin reaches Roxy to overcome queries from Roxy

### LINKED ISSUE #262: Running slush from within an existing Roxy project 'hangs'
Roxy will (try to) show the warning `Running ml new from within a Roxy project is not recommended. Continue? [yN] `, and ask to confirm if you want to continue, but the warning is not passed through by slush, nor can you enter input. So effectively it hangs without apparent reason.

We should add the same detection, or leverage the 'headless' flag that should by now be available in Roxy master..

grtjn:


grtjn:
Worked like a charm for me..

PR #315, opened 2016-02-19
Fixed #264: added error handler to less

### LINKED ISSUE #264: Less compile error should not break gulp
Not sure this still occurs in latest slush, but worth a check. Add something like:

```css
myclass {
  .label-warn;
}
```

to `main.less`, and you get this in older slush when running `gulp less`:

```
[11:30:36] Starting 'less'...

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: .label-warn is undefined in file .../ui/app/styles/main.less line no. 179
```

After which gulp terminated abnormally. If running `gulp watch`, that would break too..

grtjn:
#264

### DIFF
--- a/app/templates/gulpfile.js
+++ b/app/templates/gulpfile.js
@@ -66,10 +66,15 @@ gulp.task('plato', function(done) {
 gulp.task('styles', ['clean-styles'], function() {
   log('Compiling Less --> CSS');
 
+  var less = $.less().on('error',function(e){
+    $.util.log($.util.colors.red(e));
+    this.emit('end', e);
+  });
+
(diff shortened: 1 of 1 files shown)


grtjn:
Ran an elaborate test with this. You open `gulp serve-local`, then corrupt `main.less` by adding `.xxx;` inside one of the existing css classes, ui reloads as soon as you save and styling gets dropped. You then undo the change, and save again. ui reloads, and styling is back. The latter bit didn't happen before. Error is printed in red for bonus points..

grtjn:
Merging now..

PR #320, opened 2016-02-23
Add options for new themes, include default files to track changes

withjam:
This includes the changes to the slush file for prompting the user to select a theme.  Also the directory structure to support the themes. In this case it's a copy of the default ui so that future PRs for each theme will have a comparable history of changes.  Let me know if more/less should be in this one. #308


grtjn:
Please hold off on merging this until we release 1.0.2 and 1.0.3.. (any help with that appreciated)

withjam:
Okay, I rebased and fixed the conflict.  I've also removed the default template and tweaked the slushfile so that it doesn't overlay template directory if default is chosen.  I've kept the changes to bower.json because there is currently no method to merge in changes to the dependencies, and I don't want to keep multiple copies of the file.  There are already a bunch of unnecessary dependency modules in there so maybe there should be a new issue that involves cleaning it up and adding to it programmatically with some sort of template configs.

grtjn:
Point taken on bower.json. Let me quickly run over the files, and see if we can merge this..

app/templates/gulpfile.js
```diff
@@ -451,6 +451,12 @@ gulp.task('serve-dev', ['build'], function() {
   return serve('dev' /*env*/);
 });
 
+gulp.task('develop', ['inject','styles','less-watcher'], function() {
+    //skip the build and the browser sync
+    var nodeOptions = getNodeOptions('local');
+    $.nodemon(nodeOptions);
+});
+
```
grtjn: Sounds like a short-hand for `gulp serve-local --nosync`. Would a developer not prefer using browser sync most of the time?

app/themes/3column/ui/app/route/routes.js
```diff
@@ -0,0 +1,105 @@
...
+          stuff: function() {
+            return null;
+          }
+        }
+      })
+      .state('root.view', {
+        url: '/detail{uri:path}',
+        params: {
+          uri: {
+            squash: true,
```
grtjn: Did you take fresh copies from app/templates/ for the theme folders? I am pretty sure we removed the squash: true for instance, and there are probably several other changes..

grtjn:
Almost there..

Can you make sure you take fresh copies of app/templates/ui for the theme folders? There have been changes in there in the last two releases. Sorry, forgot to mention that earlier.

Can you also elaborate on the `develop` task? Is it much different than `gulp serve-local --nosync`? Would it help you if we would disable browser sync by default? In what way was it disturbing you?

withjam:
You're right, it still has old copies of the templates - I'll fix that.

As far as develop, I find it runs faster for me and I don't like browser sync.  For one, it launches a new browser everytime I run it and I find it easier to just save and refresh the page anyway.  I prefer a less intrusive approach.  But if gulp serve-local --nosync does the same thing then that's fine.  I didn't know that was there, I'll give it a shot.  Either way, it should probably come out of this PR.  I forgot I had added it in to be honest.

withjam:
Since each theme only has a few files of it's own, perhaps it would be easier if I just copied over the original version of those specific files and then change them as part of their specific PR so you can see the diffs?  Otherwise there will be a long list of deletions for each...

grtjn:
Do what is easiest for you..

withjam:
There's the long list of deletes that I want to avoid in the future PRs.  Instead I'll just pull over the latest from templates and copy over my changes for the diff.  That should make it easier to review the upcoming theme commits.  Let me know if this all looks ok.

grtjn:
The diff looks almost clean (except for that develop task that you wanted to remove). But only 5 files left, and nothing inside the theme dirs anymore. Are you going to restore the ones that will be touched by each UI template? I'd like to include them in this PR..

withjam:
Okay, I think this is good.  I included 2 commits in the upcoming theme branches that will includes diffs from the currently existing counterparts in templates, in addition to any newly created files for the themes themselves.  Will also make this one easier to review because there is less in it.

grtjn:
Merging. Looking forward to the other three PR's..

PR #321, opened 2016-02-24
Fixes #319 - Angular bootstrap directive changes

### LINKED ISSUE #319: Angular bootstrap directive changes
Angular bootstrap now prepends all their directive names with `uib-`. We should migrate too, and make sure all still works well.

Snippet selection seems to malfunction in latest slush, (or maybe just in UI templates)..

Audarth:
Refactored to now use ui bootstrap 0.14 and the new uib tags in the relevant html mark-up

This mostly fixes #319

### DIFF
--- a/app/templates/bower.json
+++ b/app/templates/bower.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "dependencies": {
     "angular": "~1.4.4",
-    "angular-bootstrap": "~0.13.3",
+    "angular-bootstrap": "~0.14.0",
     "angular-cookies": "~1.4.4",
     "angular-highlightjs": "~0.4.3",
     "angular-mocks": "~1.4.4",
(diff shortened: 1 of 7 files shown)


grtjn:
Looks great at first glance, thanks!

grtjn:
Nish, would it be possible to rebase this? I think you will find you can drop the `@link-color` change..

grtjn:
After rebase by Nish, anyone up for testing this? Just generate a vanilla app with this, and check if:

- pagination works/looks good
- snippet dropdown works/looks good
- same for tabs on details page
- and also login mode `modal`

ryanjdew:
@grtjn I can test.

grtjn:
@ryan321: you also wanted to help out. Would you like to take this, or can @ryanjdew have it? :)

ryanjdew:
I've tested the following and everything worked properly.

- pagination works/looks good
- snippet dropdown works/looks good
- same for tabs on details page
- and also login mode `modal`

Audarth:
Okay, I just pushed my rebased version but I tested on my local and it seemed to work fine.

grtjn:
I was wondering about typeahead, if that works or not as is. In worst case it won't work without the ml-search change, but this PR doesn't have to wait for that any longer..

PR #334, opened 2016-05-02
Profile ml security

janmichaelyu:
Using ML security instead of node server proxy to control access to user information.


janmichaelyu:
@grtjn Please review.

grtjn:
Perhaps easiest to discuss offline.. I prepared elaborate changes in #298, we might need to synch up..

grtjn:
Can you rebase?

PR #344, opened 2016-05-25
fix wiring of map markers for map theme

withjam:
Add service mlMapManager for markers array.  Updated search controller to set new markers, and root controller to watch for changes.  Also modified infoWindow to display json properties.  Note that search.controller.js will need to be modified to pull out lat/lng from wherever they are in the data.


grtjn:
Nice idea to use a service to transfer data from search ctrl to root..

PR #345, opened 2016-05-26
Fix map markers

maeisabelle:
This will fix the markers not showing on the map. Kindly review.


grtjn:
Interesting extract-path! Does that really work?? :)

If the lat/long extraction is the only change you made in the search options to accommodate the map theme, I wouldn't mind if you simply add that to all.xml. The other themes will get some extra data back, but it won't disturb anything there..

By the way, instead of the wild extract-path, you could just trim down to the property names used in the sample-data. And then make sure that the code to create the markers is aligned with those names as well. It only needs to work with the sample-data. Developers can extend from that pattern themselves..

grtjn:
By the way, thanks for looking into this!

maeisabelle:
Hi @grtjn,

Yes, it's working. :) Ok then. I'll move it to the all.xml and just retain longitude and latitude.

grtjn:
Thanks. Can you remove map.xml?

PR #347, opened 2016-05-30
Added header and footer in the Map UI theme

maeisabelle:
Kindly review the changes made. Refer to the commit message for reference. Thanks!

### DIFF
--- a/app/templates/ui/styles/main.less
+++ b/app/templates/ui/styles/main.less
@@ -124,6 +124,7 @@ nav.nav {
 
 /*footer*/
 footer {
+  position: fixed;
   bottom: 0;
   width: 100%;
   padding: 1.3em 0;
--- a/app/templates/ui/styles/main.less
+++ b/app/templates/ui/styles/main.less
@@ -124,7 +124,6 @@ nav.nav {
 
 /*footer*/
 footer {
-  position: fixed;
   bottom: 0;
   width: 100%;
   padding: 1.3em 0;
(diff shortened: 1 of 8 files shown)


grtjn:
Looks okay to me, but I'd like @jenbreese to have a look at the css as well..

maeisabelle:
Hi @grtjn and @jenbreese,

I also added changes on the detail controller so the marker for the selected documented is the only one shown when a document is clicked from the search results (same goes with the one in the Similar section). It also changes the center of the map to the position of the marker. Kindly review.

grtjn:
Looks good to me. I might have some more changes for mapManager that might interfere with this, but I'll merge that locally before pushing, and only after this PR is done..

grtjn:
@jenbreese Any comments on css code and styling from your side?

app/templates/ui/styles/main.less
```diff
@@ -124,6 +124,7 @@ nav.nav {
 
 /*footer*/
 footer {
+  position: fixed;
```
jenbreese: This is the main.less and the change needs to be in the map/theme.less

grtjn:
Merging this to make testing easier for Jen.. (and I might wanna pile up something as well..)

jenbreese:
@maeisabelle  and @grtjn I was able to test this and it worked fine. However, I didn't see the actual map controls. Are those going to be added later?

PR #348, opened 2016-05-31
Ui themes dashboard fixes

janmichaelyu:
fixing xml display tab (no more white well, added word wrap)

collapsible side bar


janmichaelyu:
@grtjn @jenbreese please review

app/themes/dashboard/ui/app/navigation/navigation.js
```diff
@@ -69,15 +72,24 @@
...
+    });
+
+    ctrl.toggleSidebar = function() {
+      var body = angular.element($document[0].body);
+      var states = { collapsed: 'sidebar-collapse', open: 'sidebar-open' };
+      var collapsed = body.hasClass(states.collapsed);
+      body.removeClass(collapsed ? states.collapsed : states.open);
+      body.addClass(collapsed ? states.open : states.collapsed);
+    };
+  }
```
grtjn: Hmm, can't we use ng-show or ng-class for this purpose, and then just toggle the value of some boolean in ctrl or $scope?

app/themes/dashboard/ui/styles/theme.less
```diff
@@ -38,4 +38,9 @@ z-index: 10;
 
 .main-header #logo {
    text-indent: 99999px;
-}
\ No newline at end of file
+}
+
+.pre-xml {
+  // background-color: white;
```
jenbreese: Can you just remove line 44 instead of commenting out?

app/themes/dashboard/ui/app/navigation/navigation.js
janmichaelyu: it looks like this is how it's done with bootstrap http://startbootstrap.com/template-overviews/simple-sidebar/

app/themes/dashboard/ui/styles/theme.less
janmichaelyu: @jenbreese removed. How does the collapsed sidebar look? Do you want a small MarkLogic image or ML logo in a similar style?

app/themes/dashboard/ui/app/navigation/navigation.js
grtjn: Yes, that is probably how it is used, but that runs on jquery directly. I'd prefer using angular, or at least hiding away jquery more..

A quick google turned up this, maybe worth a look? http://devartisans.com/articles/sidebar-navigation-using-angular

Check it out first, and see if you can find an angular way to make the sidebar work. I think it will look better. We can discuss offline before pushing too if you like..

janmichaelyu:
@grtjn How does this implementation look?

janmichaelyu:
@jenbreese Does it make sense to round the search bar and search button?

app/templates/ui/app/user/profile.controller.js
```diff
@@ -60,6 +60,7 @@
           // 'perm:sample-role': 'read',
           // 'perm:sample-role': 'update'
         }).then(function(data) {
+          toast.success('Submitted');
```
grtjn: nice..

app/templates/ui/app/error/errorInterceptor.service.js
```diff
@@ -25,6 +25,8 @@
             };
           }
           messageBoardService.message(msg);
+          var toastMsg = '<strong>' + msg.title + '</strong><p>' + msg.body + '</p>';
+          toast.danger(toastMsg);
```
grtjn: Sounds a bit like overkill, showing a message on the messageBoard as well as via toasts? But maybe I'd have to see first.. I think you can test by adding a constraint for an index that doesn't exist. Search will then break, and should normally trigger the messageboard..

app/themes/dashboard/ui/index.html
```diff
@@ -19,8 +19,8 @@
     <!-- endbuild -->
 
   </head>
-  <body ng-app="app" class="skin-black">
-
+  <body ng-app="app" class="skin-black sidebar-mini" ng-controller="SidebarController" sidebar-directive="sidebarState">
+    <toast></toast>
```
grtjn: Kinda wondering if this needs to be in index.html. If it could be in root.html, then you could use root controller.

Alternatively, how about just sticking sidebarState in $rootScope. Having this ng-controller on body really doesn't look right. Sounds like opening a hole for potential side-effects..

app/themes/dashboard/ui/app/sidebar/sidebar.directive.js
```diff
@@ -0,0 +1,20 @@
...
+                element.removeClass(states.collapsed);
+                return;
+              }
+              element.removeClass(states.open);
+              element.addClass(states.collapsed);
+            });
+          }
+        };
+    });
+}());
```
grtjn: Can't you achieve the same with ng-class on the body element?

grtjn:
Still trying to wrap my head around that sidebar stuff.. By the way, does this include both the toggle for the left-hand sidebar, as well as the toggle for the right-hand sidebar?

app/templates/bower.json
```diff
@@ -20,7 +20,10 @@
     "ng-json-explorer": "8c2a0f9104",
     "vkbeautify-wrapper": "*",
     "highcharts": "^4.2",
-    "angular-google-maps": "2.3.2"
+    "angular-google-maps": "2.3.2",
+    "angular-component.js": "^0.0.8",
```
grtjn: I don't think you need angular-component for the sidebar..

app/themes/dashboard/ui/styles/theme.less
```diff
@@ -43,3 +43,11 @@ z-index: 10;
 .pre-xml {
   background-color: #ecf0f5;
 }
+
+.nav-pills>li>a {
+  border-radius: 4px !important;
```
jenbreese: Please remove the rounded corners. This design is supposed to have square buttons.

app/themes/dashboard/ui/styles/theme.less
```diff
@@ -43,3 +43,11 @@ z-index: 10;
...
+}
+
+.nav-pills>li.active>a {
+  background-color: #384246 !important;
```
jenbreese: This line should be removed. There are multiple themes that can be chosen and this !important will cause a lot of trouble.

app/themes/dashboard/ui/styles/theme.less
```diff
@@ -38,4 +38,16 @@ z-index: 10;
...
-}
\ No newline at end of file
+}
+
+.pre-xml {
+  background-color: #ecf0f5;
+}
+
+.nav-pills>li>a {
+  border-radius: 4px !important;
```
jenbreese: The design is for square buttons and not rounded ones. This should be removed. Thanks!

app/themes/dashboard/ui/styles/theme.less
grtjn: Well, this bit is in theme.less, so won't interfere, but I agree we should try to avoid !important if possible.

By the way, if the !important was necessary to override something from main.less, I will probably be moving the contents of main.less into a separate import that will be loaded *before* theme.less. I think that should make it easier to override our default designs..

grtjn:
@janmichaelyu A thought struck me. What if we move all dashboard specific stuff in index.html, and put that in root.html, you will have root controller within reach, and will be able to go through that to make things work. Saves you the extra sidebar controller for starters.

Can you try that, and see if for instance the UI still looks the same? It does mean moving skin-black from body to an inner div, but I hope that works..

app/themes/dashboard/ui/index.html
janmichaelyu: Got it. I was having trouble working with the sidebar-open and sidebar-collapsed css classes in root.html.

app/themes/dashboard/ui/styles/theme.less
janmichaelyu: I had to use !important to override a theme (adminLTE.less) that was imported inside theme.less. Based on another comment, I'll remove the rounding for the buttons.

jenbreese: Yes, @grtjn and I tested it today and I'd prefer we didn't change the color of those pill buttons. I think that should be a choice made by the engineer using the template.

janmichaelyu:
@grtjn How does this look? Even moving the header into root the css class needs to apply to the ui-view as well. If I have to move the header into root I think I also have to make the ui-view parts to use the sidebar css classes.

grtjn:
I don't think this needs to be complex. Try the following:

- Move the part from your current index.html from toast until ui-view and literally insert that into root.html
- The same for the part after ui-view in index.html till the build comments

This should have no UI impact, but perhaps the toggle gets corrupted. For that we do:

- Change the line with the div with the sidebar controller into: `<div class="wrapper skin-black sidebar-mini" ng-class="{'sidebar-open': !!ctrl.navService.showSidebar, 'sidebar-collapse': !ctrl.navService.showSidebar}">`
- Change the ng-click of the sidebar-toggle to read: `ng-click="ctrl.navService.toggleSidebar()"`
- And then simplify the toggleSidebar function in the nav service to:

```
      service.showSidebar = true;
 
      service.toggleSidebar = function () {
        service.showSidebar = !service.showSidebar;
      };
```

That should be enough to get the sidebar toggle working. You can then:

- Undo TopNavCtrl changes
- Remove sidebar controller

By the way:

- Please remove detail controller. (We accidentally changed the generic one, but that has been fixed already..)
- Please also remove the theme.less change for `.nav-pills>li.active>a` with the !important color change. (request on behalf of Jen)

Last but not least:

- Can you move that toast change in the profile controller to the one in app/templates to apply that generically.
- Add a similar toast to the create controller in app/templates
- Add `<toast>` to the root.html in app/templates (and to the other theme root.html's too perhaps)
- Do the toasts still work if you move the `<toast>` from before the `<div>` to inside it?

janmichaelyu:
Toasts are working fine in Create and Profile update in dashboard and default theme.

app/templates/bower.json
grtjn: Please remove angular-component dependency..

app/templates/ui/index.html
```diff
@@ -20,7 +20,7 @@
 
   </head>
   <body ng-app="app">
-
+    <div><toast></toast></div>
     <ui-view></ui-view>
 
     <!-- build:js js/lib.js -->
```
grtjn: I think the `<toast>` tag should be in root.html. Which means you will need to check the other themes as well..

app/themes/dashboard/ui/app/navigation/navigation.js
```diff
@@ -66,18 +69,29 @@
...
+        service.showSidebar = !service.showSidebar;
+      };
+
       return service;
     }
 
-    TopNavCtrl.$injector = ['$scope', 'userService'];
-    function TopNavCtrl($scope, userService) {
-      var ctrl = this;
+  TopNavCtrl.$injector = ['$scope', 'userService', '$document', 'loginService'];
```
grtjn: $document is unused..

app/themes/dashboard/ui/app/navigation/navigation.js
```diff
@@ -66,18 +69,29 @@
...
+    $scope.$watch(userService.currentUser, function(newValue) {
+      ctrl.currentUser = newValue;
+    });
+
+    ctrl.logout = function() {
+      loginService.logout();
+    };
+  }
 
 })();
```
grtjn: I was adding userService to the root controller already in the other PR. How about exposing the loginService as well, and just getting rid of TopNavCtrl altogether. You can expose user and loginService in the generic root controller to make them generally available..

app/themes/dashboard/ui/app/sidebar/sidebar.controller.js
```diff
@@ -0,0 +1,22 @@
...
+      if (newVal) {
+        $scope.sidebarClass = 'wrapper skin-black sidebar-mini sidebar-open';
+      } else {
+        $scope.sidebarClass = 'wrapper skin-black sidebar-mini sidebar-collapse';
+      }
+
+    });
+
+  }
+}());
```
grtjn: I think sidebar controller is fully redundant now..

grtjn:
Hi @janmichaelyu, thanks for the changes. I have a few last minute change requests, see inline comments..

Re changing colors of pills: Jen suggested just leaving them as they are, it is how AdminLTE is supposed to look.. :-)

janmichaelyu:
@grtjn Got it. Made the changes you requested and also removed the links from the topnavbar.

grtjn:
Thanks for the quick turnaround! Let me merge this, so I can rebase my PR on top of it..

PR #349, opened 2016-06-02
Refined map theme (and a few other small changes)

grtjn:
- wrapped up map theme improvements (merged with Mae)
- fixed issue of watching less files
- restored original detail.ctrl, moved changed one into themes/map/
- removed redundant 3-column/index.html (was identical to default one)


grtjn:
Merging to make testing easier..

PR #350, opened 2016-06-02
Various general code improvements, and a map theme bugfix

grtjn:


grtjn:
@jenbreese Could you glance over this PR to see if it includes all as discussed? Feel free to merge it as well if you are allowed to, so that you can run actual tests with it. I had a quick look, and only default theme looked a little degraded, but maybe you spot some more things that aren't right..

grtjn:
Closing this PR in favor of #353, which is essentially just a rebased version of this one..

PR #352, opened 2016-06-06
Added pm2 config generation

### LINKED ISSUE #351: pm2 integration
As a follow-up to my pm2 work a few weeks ago I've hacked up 2 new tasks, `gulp init-process` and `gulp init-ecosystem`. These are based on the `gulp init-local` task and I've also put their equivalents into the slushfile for when a new project is created as well as adding ecosystem.json to .gitignore

The idea so far is that process.json can be run with `pm2 start process.json` and ecosystem.json will contain any credentials and be used for deployment with `pm2 deploy target setup` and `pm2 deploy target`. ecosystem.json contains a sample-target but it's stripped of any real details and is meant to act as reference for creating real targets.

This is just a bit of a starting point so let me know what you think and I'll make changes accordingly.

Audarth:
resolves #351


grtjn:
Maybe add npm install pm2 to the pre-requisites? And a section for the README that explains how to use it and how it works?

Audarth:
I believe pm2 should always be installed globally so I don't think putting it into package.json is necessary. 

I'll update the README now but I think I'll leave it simple and just put in the stop/start app and deploy commands, there's quite a few other commands though so it might be good for me to make a wiki page to go into more detail.

grtjn:
yeah, I didn't mean adding it to package.json. README is good, wiki even better. As long as what is in this PR contains/describes all minimal details to use and understand..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
 the part that handles showing your data, can not only handle JSON, but also
 XML, Binary, and Text data out of the box.
+
+## Running with PM2
+
+To run you must install pm2 with:
+
+`npm install pm2 -g`
+
+Make sure you have the file `process.json` if it's missing then run:
```
grtjn: Since `process.json` is not a default name for pm2, I vote calling it `boot.json`, since that is a pattern that should be relatively familiar to slush users.

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+`gulp init-process`
+
+This file will also include any config your node layer needs eg. mlHost, mlPort, nodePort, mlPassword etc.
+
+You can then launch the app using pm2 with:
+
+`pm2 startOrRestart process.json`
```
grtjn: On servers running the pm2 daemon with a pm2 user, make sure to pm2 start/save as pm2 user (`sudo su pm2`)

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+You can also launch the app using different environment configuration (env name prefixed with "env_") using:
+
+`pm2 startOrRestart process.json --env build`
```
grtjn: Please take local, dev, and prod as example instead of build. Prod and dev both require 'NODE_ENV=build'..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+To view all running pm2 applications use:
+
+`pm2 list`
```
grtjn: See above, run as pm2 users.. (mention pm2 kill if you accidentally launch it under your own account). Or should we mention `sudo -u pm2 pm2 list` instead to prevent mistakes?

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+## Deploying with PM2
+
+To deploy, make sure both your local machine and your remote server/deployment target have pm2 installed with:
+
+`npm install pm2 -g`
+
+Make sure you have the file `ecosystem.json` if it's missing then run:
+
+`gulp init-ecosystem`
```
grtjn: how about `add-deploy-target` (which creates an ecosystem.json if needed). It can question for key, username, git url and provide defaults leveraging local into (local user, local git details)..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+`ecosystem.json` will include any config as well as any deployment config along with your credentials. It is git ignored by default and we recommend you avoid commiting it and instead put any repository appropriate configuration in `process.json`
+
+Using the sample target in ecosystem.json as a guide, set up a target to deploy to. Your configuration must have an ssh key, remote server user, git repository potentially with credentials and a valid path that isn't being used.
+
+First initialize the target host:
+
+`pm2 deploy ecosystem.json TARGETNAME setup`
```
grtjn: skip the ecosystem.json, consider replacing TARGETNAME with `[local | dev | prod]`..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+Then deploy new code with:
+
+`pm2 deploy ecosystem.json TARGETNAME`
```
grtjn: `pm2 deploy prod` required after setup to trigger post-deploy hook..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+**Hint**, If you do not specify `ecosystem.json` pm2 automatically looks for it in your current directory
+
+If you want to deploy without commiting your current code use `--force`:
+
+`pm2 deploy TARGETNAME --force`
+
+You can also run commands on the server with `exec` such as starting the app:
+
+`pm2 deploy ecosystem.json TARGETNAME exec "pm2 startOrRestart process.json"`
```
grtjn: Don't! You need to run that with pm2 user, and you can't do sudo -u pm2 as you won't be able to type in pwd..

app/templates/README.mdown
```diff
@@ -69,3 +69,67 @@ Or on Windows:
...
+
+For more information use:
+
+`pm2 deploy help`
```
grtjn: Don't forget about installing pm2 service on server, doing pm2 start as pm2 user, pm2 save after each deploy setup + start (necessary?).

Also mention pm2 list, pm2 logs, pm2 monit..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+    var configJSON = pm2JSON();
+
+    var configString = JSON.stringify(configJSON, null, 2) + '\n';
+    fs.writeFileSync('process.json', configString, {
+      encoding: 'utf8'
+    });
+  } catch (e) {
+    console.log('failed to write process.json: ' + e.message);
+  }
+});
```
grtjn: We discussed adding a fixed process.json/boot.json to templates, so you don't really need this, particularly since we noticed only app name will be different for each app..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+/**
+ * Creates a sample ecosystem.json
+ */
+gulp.task('init-ecosystem', function() {
+  log('Creating ecosystem.json sample document with values drawn from gulp.config.js');
+  try {
+    var configJSON = pm2JSON();
+
+    configJSON.deploy = {
+      'sample-target': {
```
grtjn: use local/dev/prod as targetname..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+    'ML_HOST': config.marklogic.host,
+    'ML_PORT': config.marklogic.httpPort,
+    'ML_APP_USER': config.marklogic.user,
+    'ML_APP_PASS': config.marklogic.password
+  };
+
+  configJSON.apps = [apps];
+
+  return configJSON;
+}
```
grtjn: We came up with this file instead together:

```
{
  "apps": [
    {
      "name": "@sample-app-name",
      "script": "./node-server/node-app.js",
      "watch": true,
      "env": {
        "NODE_ENV": "build"
      }
    }
  ]
}
```

Just pondering if it makes sense to have three envs in there, for local, dev and prod. And which should be default?

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+ * Creates a sample ecosystem.json
+ */
+gulp.task('init-ecosystem', function() {
+  log('Creating ecosystem.json sample document with values drawn from gulp.config.js');
+  try {
+    var configJSON = pm2JSON();
+
+    configJSON.deploy = {
+      'sample-target': {
+        'key': '/path/to/key',
```
grtjn: ~/.ssh/id_rsa, but then expanded to abs path..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'user': 'USERNAME',
```
grtjn: current user name

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'host': ['HOSTNAME'],
+        'ref': 'origin/master',
```
grtjn: figure out if we can detect that from local git details.. git status perhaps?

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'repo': 'https://GIT_USERNAME:GIT_PASSWORD@marklogic.unfuddle.com/git/REPO_NAME/',
```
grtjn: I don't like having git user and git pwd exposed like this. I think we should require people to setup their ssh key such that a git clone from github or unfuddle will that instead. You only need to make sure the fingerprint is known, and it will work (we tried)..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'path': '/space/projects/PROJECT_FOLDER/',
```
grtjn: /space/projects/@sample-app-name/.

Note: /space/projects needs to be writable for the current user on the server..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'pre-deploy-local': 'echo "an echo was executed on your local machine"',
```
grtjn: remove this..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'post-deploy': './ml local bootstrap --ml.password=SERVER_PASSWORD; ./ml local deploy modules --ml.password=SERVER_PASSWORD; ./ml local deploy content -password SERVER_PASSWORD; npm install; bower install; gulp build',
```
grtjn: Change this into "post-deploy": "npm install; bower install; gulp build --ignoreErrors". Roxy has good means to ask for pwd, and deploy remotely. Let's stick with that, adding the pwd in here repeatedly is both annoying and unsafe..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+        'env': {
+          'NODE_ENV': 'production'
+        }
```
grtjn: env is not needed here..

app/templates/gulpfile.js
```diff
@@ -231,6 +231,91 @@ gulp.task('init-local', function() {
...
+  }
+});
+
+/**
+ * Creates a sample ecosystem.json
+ */
+gulp.task('init-ecosystem', function() {
+  log('Creating ecosystem.json sample document with values drawn from gulp.config.js');
+  try {
+    var configJSON = pm2JSON();
```
grtjn: ecosystem.json is gitignored, there is not much use adding apps section in it..

app/templates/gulpfile.js
grtjn: Note: it assumes you provide the essential details via local/dev/prod.json via the mechanism we created before. (I have improved that further in PR #298)

app/templates/gulpfile.js
grtjn: Adding fingerprints, needs to be done by every uses that wants to deploy, on each of the machines they want to deploy.

Github: ssh into the server, and run `ssh -T git@github.com`, accept fingerprint
Unfuddle: no elegant way, ssh into the server, and run git clone against any git repo on unfuddle, accept fingerprint, terminate the git clone once past that stage, remove any subdir it created..

app/templates/_gitignore
```diff
@@ -5,6 +5,7 @@ ui/fonts
 ui/styles/*.css
 deploy/local.properties
 local.json
+ecosystem.json
```
grtjn: Not sure this really needs to be in here. There will be no pwds in there, if assuming ssh keys, and not running Roxy post-deploy. There might be a username, but pm2 deploy is only used to deploy to dev or prod, and that is usually done by just one person. If someone else needs to take over, he/she can edit the file temporarily. At least server name and git location are already entered, so it would save colleagues time..

app/templates/README.mdown
grtjn: Additional note on pm2 start: must be run from code root, as process/boot.json has a relative path to node-server/node-app.js..

app/templates/_gitignore
Audarth: Hmm I think if it's repository safe devs can put the deploy target in boot.json. It would also just be fairly useful for people to have somewhere they can put their user specific info that's safe from git, just in case they need to put in credentials.

app/templates/gulpfile.js
```diff
@@ -224,19 +235,118 @@ gulp.task('init-local', function() {
...
+        gulp.src(ecosystem)
+          .pipe($.jsonEditor(function(json) {
+            json.deploy[answers.targetName] = {
+              'key': answers.key,
+              'user': answers.username,
+              'host': [answers.hostname],
+              'ref': 'origin/' + answers.branch,
+              'repo': answers.gitUrl,
+              'path': answers.folder,
+              'post-deploy': 'npm install; bower install; gulp build'
```
grtjn: Was wondering if we should use `&&` instead of `;` between the post-deploy commands. That will make post-deploy stop after npm install if that happens to fail.

grtjn:
Can you rebase. We also still need to clean up those whitespace changes a bit. Perhaps go over it together?

Audarth:
Sure

app/templates/gulpfile.js
```diff
@@ -294,41 +436,49 @@ gulp.task('optimize', ['inject', 'test'], function() {
...
+    restore: true
+  });
 
   var templateCache = config.temp + config.templateCache.file;
 
   var combined = gulp
     .src(config.index)
     .pipe($.plumber())
     .pipe(inject(templateCache, 'templates'))
-
```
grtjn: Is it possible to re-instate the empty lines in this bit?

app/templates/gulpfile.js
```diff
@@ -294,41 +436,49 @@ gulp.task('optimize', ['inject', 'test'], function() {
...
-  var jslibFilter = $.filter('**/' + config.optimized.lib, {restore: true});
+  var cssFilter = $.filter('**/*.css', {
+    restore: true
+  });
+  var jsAppFilter = $.filter('**/' + config.optimized.app, {
+    restore: true
+  });
+  var jslibFilter = $.filter('**/' + config.optimized.lib, {
+    restore: true
+  });
```
grtjn: Can you put the `{ restore: true }` in a var, so that this will look more dense?

slushfile.js
```diff
@@ -240,13 +241,17 @@ function configRoxy() {
       'user=' + settings.marklogicAdminUser + '\n' +
       'password=' + settings.marklogicAdminPass + '\n';
 
-    fs.writeFileSync('deploy/local.properties', localProperties, {encoding: 'utf8'});
+    fs.writeFileSync('deploy/local.properties', localProperties, {
+      encoding: 'utf8'
+    });
```
grtjn: `{ encoding: 'utf8' }` is used in a few places, put in a var for denser code as well?

app/templates/gulpfile.js
Audarth: Hmm my beautifier doesn't really like it when I try and doesn't align all the methods properly

grtjn: Ah right, we looked at that. Never mind then..

slushfile.js
```diff
@@ -419,39 +501,53 @@ gulp.task('init', ['checkForUpdates'], function (done) {
...
+            .pipe(rename(function(file) {
+              // change _foo to .foo
+              if (file.basename[0] === '_') {
+                file.basename = '.' + file.basename.slice(1);
+              }
+
+            }))
+            .pipe(replace('@slush-version', pkgSettings.version.trim(), {
+              skipBinary: true
+            }))
```
grtjn: can you put skipBinary in a var as well, like encoding?

grtjn: Just this last formatting thing, and I think we can merge..

PR #353, opened 2016-06-06
Map fixes

grtjn:


grtjn:
Merging this now, going to pile up some more..

Issue #332, opened 2016-04-22
Avoid /v1 in proxy.js


grtjn:
The proxy function has a hard-coded '/v1', but it is better to use `req.baseUrl` instead. That also makes it easier to reuse the same proxy lib to proxy '/vlatest' for instance, or add-hoc proxies like '/reports'(/my/report.xqy)..

grtjn:
A fix for this was included in 1.1.0 already..

PR #354, opened 2016-06-06
Minimizing diffs between ui-themes

grtjn:
and some other small last refinements..


grtjn:
A few more things coming up..

grtjn:
This included a fix for #332 as well..

PR #358, opened 2016-06-14
Fixing infoWindow, and improving display of random json in it

grtjn:


grtjn:
Merging now to allow rebasing the other ui-themes PR on top of it..

PR #367, opened 2016-06-23
Adopting Angular 1.5 and refactoring existing code to use Components

### LINKED ISSUE #357: forward-compatibility with angular 2
we should start porting directives to components (including the "Community Add-ons"), for forward-compatibility with angular 2. The `angular-component` polyfill brings support back to angular 1.3.

Audarth:
Still need a bit more work but starts to resolve #357


grtjn:
Could you add the angular-component polyfill for angular 1.4, and then change back minimum version of angular parts to 1.4.4, but allow 1.4.4 - 1.9? I think you can do that with using `^1.4.4` instead of `~1.4.4`. Not sure how easy it will be to test that, but I like a bit of backwards-compat. No need to upgrade dependency min-versions more than really necessary..

Audarth:
Hmm I've thought about this but I do think it would probably be better to just go with 1.5 in slush and reserve the use of the polyfill for older projects where upgrading is difficult but we want to utilize components we've written.

The main advantage of having a min version of 1.4 and using the polyfill is that we have a bit more version flexibility.

But 1.5 is probably the last major angular 1 version anyway and I think all the angular libraries will converge on it and make it the most popular version until angular 2 get's more mature.

It's also overall simpler to just go with 1.5 rather than mixing the 1.5 code with 1.4 code and we can take advantage of other features not included in the component polyfill. We'll also probably be less likely to have weird bugs if we avoid mixing versions.

grtjn:
Can you name a few of such features not available with polyfill? Will we need those? Just trying to get a clear picture..


Sent from Samsung Mobile

-------- Oorspronkelijk bericht --------
Van: Audarth <notifications@github.com> 
Datum: 28-06-2016  03:56  (GMT+01:00) 
Aan: marklogic/slush-marklogic-node <slush-marklogic-node@noreply.github.com> 
Cc: Geert <geert.josten@xs4all.nl>,Comment <comment@noreply.github.com> 
Onderwerp: Re: [marklogic/slush-marklogic-node] Adopting Angular 1.5 and refactoring existing code to use Components (#367) 
 
Hmm I've thought about this but I do think it would probably be better to just go with 1.5 in slush and reserve the use of the polyfill for older projects where upgrading is difficult but we want to utilize components we've written.

The main advantage of having a min version of 1.4 and using the polyfill is that we have a bit more version flexibility.

But 1.5 is probably the last major angular 1 version anyway and I think all the angular libraries will converge on it and make it the most popular version until angular 2 get's more mature.

It's also overall simpler to just go with 1.5 rather than mixing the 1.5 code with 1.4 code and we can take advantage of other features not included in the component polyfill. We'll also probably be less likely to have weird bugs if we avoid mixing versions.

grtjn:
Note: we can always reconsider in later stages. Whatever we pick now, we can always come back to decisions, and adjust later on..

Audarth:
https://docs.angularjs.org/guide/migration

The main one is the improved support for es6 such as using classes instead of controllers which allows you to use syntax more similar to angular 2 as seen in this [styleguide](https://github.com/toddmotto/angular-styleguide)

Audarth:
I guess it kind of depends on how much we want to do for forward compatibility, if we want to just refactor directives into components then the polyfill is enough for now but if we want to also start using es6/es2015 and writing more angular 2 style code then 1.5 is necessary.

grtjn:
I like the idea of multiple smaller steps, but maybe worth pushing the limit a bit as well to get a better idea of how big the full impact for Angular 2 would as i have no idea to be honest..

Audarth:
So with these latest commits everything is now refactored to use components instead of directives, could maybe get someone to verify that everything still works as it should.

I might now create another branch and start to experiment with mixing in es6 code, I could put that in a separate pull request.

grtjn:
Separate branch and pr for the es6 stuff makes sense. I need to merge my default-user and your pm2 PR before I can sensibly review this one though..

grtjn:
Any luck with rebasing yet? :)

Have you considered rewriting code to use one-directional bindings?

Audarth:
I mixed some of the commits from the pm2 stuff in here so that's a bit tricky, combined with merging in the changes from the last few pull requests since there's a ton of conflicts now which will require some more coding from me

grtjn:
I notice you got rid of the login/logout events. userService no longer depends on loginService. Does that work?

grtjn:
We also need to synch up a lot, made quite some changes. Lets get in touch offline, and maybe work on it together?

grtjn:
As discussed offline: this PR has in need of a rebase once more. But perhaps changes have become too large, so reimplementing might be easier. Consider opening a new PR against cards-theme branch in which you re-implement the essential changes. In that case, please only touch files that really change.. ;-)

PR #371, opened 2016-07-19
Utilizing ES6, webpack and babel with angular 1.5

### LINKED ISSUE #370: ES6, webpack and forward compatibility
Creating a new issue since this is a bit different from #357 

There's a number of good projects based around creating an angular 1.5 project that's forward compatible with angular 2 and leverages newer technologies like webpack and ES6.

It might be good to move in this direction, I've already created a bit of a prototype based on [this project](https://github.com/AngularClass/NG6-starter) and it seems to work quite well. the code becomes a lot more modular thanks to ES6 and webpack builds the interface to be less than 1 MB compared to around 2 MB for the current default slush running from serve-prod.

I'll put in a pull request though it's definitely far from complete as I haven't supported the themes, intend to convert all of the code to use es6 syntax and need to clean up a lot of code that's no longer used.

Audarth:
Starts to resolve #370


grtjn:
Looks very promising already, but could you reopen this PR against the new es6 branch? We can then merge this PR, and work/test on it in parallel..

Audarth:
Done! I've also pushed a bunch of changes to get the latest master changes working and improve the unit tests.

PR #372, opened 2016-07-20
Added sudo and amended some of the commands for removing services

Audarth:
Most of the commands for removing services need sudo, I've also fixed some syntax errors I missed


app/templates/INSTALL.mdown
```diff
@@ -112,39 +112,41 @@ When converting a project to pm2 you should create an `ecosystem.json` file with
...
-`chkconfig -del {appname}-watch`
+`sudo chkconfig --del {appname}-watch`
 
 Then delete the scripts
 
-`rm /etc/init.d/{appname}`
+`sudo rm /etc/init.d/{appname}`
 
-`rm /etc/{appname}`
+`sudo rm -r /etc/{appname}`
```
grtjn: Don't use `-r`, it is a symbolic link, not a complete directory..

app/templates/INSTALL.mdown
```diff
@@ -112,39 +112,41 @@ When converting a project to pm2 you should create an `ecosystem.json` file with
...
 
 And for the watch script:
 
-`rm /etc/init.d/{appname}-watch`
+`sudo rm /etc/init.d/{appname}-watch`
+
+`sudo rm -r /etc/{appname}-watch`
```
grtjn: same here..

grtjn:
I happen to notice we have left etc/ in the slush template. Didn't we plan to remove it to prevent confusion?

Audarth:
Yeah, I think it would be best

app/templates/INSTALL.mdown
Audarth: Hmm I had to use -r when removing the config for obi but it looks like just rm worked for alerts so it was probably just obi being a bit odd, I'll change it now

grtjn:
Looking good, thanks, merging now!

PR #374, opened 2016-08-05
Add triples to sample data

patrickmcelwee:
This commit adds the following triples to each sample document:

    <docUri> <rdfs:label> <doc/name> .
    <docUri> <rdf:type> <foaf:Person> .

and three triples pointing to 3 random documents:

    <docUri> <foaf:knows> <someRandomDocUriFromThisSampleSet>

There is also a single additional triple in the first doc ('data-0.json') that establishes a label for `foaf:knows`. This would normally go into an ontology, but it seemed easiest just to embed it into a document rather than introduce that much complexity. This additional triple is needed for a nice OOTB display when using [ml-visjs-graph-ng](https://github.com/patrickmcelwee/ml-visjs-graph-ng)

These triples were inserted into the original sample docs using [this sample-triple-generator library](https://github.com/patrickmcelwee/sample-triple-generator), in case that is a helpful starting point for any future additions.


grtjn:
Sounds great!

grtjn:
After taking a closer look at the data, I notice the docUri subject is a bit fragile, as it must match the ultimate database uri I guess. What difference would it make if you'd use _id value instead of database uri as subject?

patrickmcelwee:
That's interesting ... I did this while creating [ml-visjs-graph-ng](https://github.com/patrickmcelwee/ml-visjs-graph-ng). That triple-visualization library has a function that will go to the detail page for a node when right-clicked, assuming the RDF URI of the node is a MarkLogic URI. So it made sense to set it up this way, to demonstrate that.

Do you think the MarkLogic URIs for the sample will change often?

grtjn:
Not really, but just change the uri prefix on import and it breaks. Wondering if there would be a more robust way of getting the doc uri, dynamically for instance.. not via sparql I thought though.. and doing a transform on ingest is perhaps a bit too much as well..

Maybe i'm overthinking this..

patrickmcelwee:
I see what you mean. We *could* write some server-side code that created these triples based on docs already loaded into the database (or a transform, as you say). But, since this is just sample data, and we provide them with the correct configuration, I am inclined to say we should just hard-code the docUris for now - until someone points out that it is a terrible idea.

grtjn:
Yeah, agree..

I'll see if i can do a quick sanity check and then merge..

By the way, would be interested in seeing that vis component. Do you have a live demo on github.io or elsewhere?


Sent from Samsung Mobile

-------- Oorspronkelijk bericht --------
Van: Patrick McElwee <notifications@github.com> 
Datum: 18-08-2016  20:18  (GMT+01:00) 
Aan: marklogic/slush-marklogic-node <slush-marklogic-node@noreply.github.com> 
Cc: Geert <geert.josten@xs4all.nl>,Comment <comment@noreply.github.com> 
Onderwerp: Re: [marklogic/slush-marklogic-node] Add triples to sample data (#374) 
 
I see what you mean. We could write some server-side code that created these triples based on docs already loaded into the database (or a transform, as you say). But, since this is just sample data, and we provide them with the correct configuration, I am inclined to say we should just hard-code the docUris for now - until someone points out that it is a terrible idea.

patrickmcelwee:
I don't have a live demo, but would like to. Do you have any tips on how to do that when it depends on MarkLogic running in the background? Do you mock out the MarkLogic calls? ... I was going to try to use it in a demo, but something in isolation on github.io would be nice

grtjn:
I faked search responses for ml-histogram-ng, but you feed that with data, not a connection.

Maybe talk offline to discuss ideas?


Sent from Samsung Mobile

-------- Oorspronkelijk bericht --------
Van: Patrick McElwee <notifications@github.com> 
Datum: 18-08-2016  21:42  (GMT+01:00) 
Aan: marklogic/slush-marklogic-node <slush-marklogic-node@noreply.github.com> 
Cc: Geert <geert.josten@xs4all.nl>,Comment <comment@noreply.github.com> 
Onderwerp: Re: [marklogic/slush-marklogic-node] Add triples to sample data (#374) 
 
I don't have a live demo, but would like to. Do you have any tips on how to do that when it depends on MarkLogic running in the background? Do you mock out the MarkLogic calls? ... I was going to try to use it in a demo, but something in isolation on github.io would be nice

grtjn:
This has waited long enough, moving forward..

PR #380, opened 2016-08-18
Hide 'Edit' and 'Delete' when `disallowUpdates: true`

patrickmcelwee:
Previously, the buttons still appeared, but resulted in a blank page for 'Edit' and an unhandled server error for 'Delete'.

Note that this does not remove the 'Edit' and 'Delete' buttons for a Guest (just as the Create button is not removed for a guest) - this would be nice, but there is currently no different in `ctrl.currentUser` between a logged-in app-user and a guest, so no way I can see to check in the template.

This requires adding the userService and $scope as additional dependencies in the generated detail controller.

It seems that only the map theme has an override for the detail controller. I made the changes there as well.

I tested this against the default template and the map theme.

### DIFF
--- a/app/templates/ui/app/detail/detail.controller.js
+++ b/app/templates/ui/app/detail/detail.controller.js
@@ -4,8 +4,9 @@
   angular.module('app.detail')
   .controller('DetailCtrl', DetailCtrl);
 
-  DetailCtrl.$inject = ['doc', '$stateParams','MLRest', 'ngToast','$state'];
-  function DetailCtrl(doc, $stateParams, MLRest, toast, $state) {
+  DetailCtrl.$inject = ['doc', '$stateParams','MLRest', 'ngToast',
+                        '$state','$scope','userService'];
+  function DetailCtrl(doc, $stateParams, MLRest, toast, $state, $scope, userService) {
     var ctrl = this;
 
     var uri = $stateParams.uri;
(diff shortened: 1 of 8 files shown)


grtjn:
Thanks for spotting this, I totally forgot to test with those flags..

grtjn:
Looks pretty clean at first glance..

patrickmcelwee:
Yeah, it's pretty tough to test now, with all the various possible configurations. Have we ever considered writing tests that generate applications and run tests against them on a CI server somewhere? They might be tough tests to write, but it is also going to be tough to manually test all this going forward.

grtjn:
Generating all UI flavors, and running gulp build on them should be pretty straight-forward. Not sure how easy one can fake user input though, to skip through the slush wizard..

grtjn:
But the gulp build would not capture these kind of issues, or perhaps one could, but it will be a pretty touch test to write..

patrickmcelwee:
Yeah, in a couple other projects, I've written end-to-end tests that make assertions on which elements are in the DOM, their values, etc; and also drive user interaction through the browser. You're right: they are often brittle (and slow - I would put them in a separate, occasionally run suite). On the other hand, it reduces the need for error-prone manual checking. Trade-offs. :-|

grtjn:
This has waited long enough. I'll republish the slush demo app's on short notice, that will allow doing sanity checks.

FYI: I have ramped up an internal CI server that does gulp build on all four themes on a daily basis..

Issue #389, opened 2016-09-21
Gulp build breaks with old nodejs (0.10)


grtjn:
You might see:

```
[15:48:54] gulp-inject 1 files into index.html.

/Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/node_modules/svgo/plugins/cleanupIDs.js:123
    for (var ID of referencesIDs) {
                ^^
SyntaxError: Unexpected identifier
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/node_modules/svgo/lib/svgo/config.js:99:33
    at Array.map (native)
    at preparePluginsArray (/Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/node_modules/svgo/lib/svgo/config.js:64:20)
    at module.exports (/Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/node_modules/svgo/lib/svgo/config.js:32:28)
    at new module.exports (/Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/node_modules/svgo/lib/svgo.js:21:19)
    at /Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss-svgo/dist/index.js:91:16
    at Object.creator [as postcssSvgo] (/Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/node_modules/postcss/lib/postcss.js:150:35)
    at /Users/gjosten/Projects/marklogic_kpn-nectar/node_modules/gulp-cssnano/node_modules/cssnano/dist/index.js:276:40
    at Array.forEach (native)
[...]

PR #392, opened 2016-09-26
391 generator options

grtjn:
#391, #389


grtjn:
Tested locally, seems to run smooth..

PR #413, opened 2016-11-15
Fine-tuned dashboard charts, and fixed node-server proxy

grtjn:


grtjn:
Ran elaborate tests locally, merging now..

Issue #410, opened 2016-10-31
Difficulty running MLCP import under Windows


grtjn:
Report from the field that running the MLCP import step as documented in the README results in Java help being printed on the console, from both cmd and powershell. The README should also elaborate on properly setting mlcp-home..

grtjn:
Java help is printed if mlcp-home is mis-configured. Both README and WINDOWS docs have a section on mlcp-home that hopefully prevents new people from hitting this..

grtjn:
Thnx @bluetorch for digging into this!

PR #426, opened 2016-11-23
Adding documentation for the proper use of MLCP in Windows.

bluetorch:
#410 Windows paths use the backslash, which is a reserved character in Java.  Documentation added to show a windows user how to configure MLCP for use with Roxy.


grtjn:
Thnx! Do you also observe the same strange effect that java help is being printed if you invoke MLCP via Roxy?

bluetorch:
If you do not double the backslash it does print the java help.

grtjn:
Excellent, that makes a lot of sense. Jar not found, class not found, Java doesn't understand the call you make, and prints a not very helpful usage message instead of a more to the point error message..

PR #435, opened 2016-12-21
Do not add padding to .row in the dashboard theme

patrickmcelwee:
In Bootstrap, .row has no padding (and in fact has a negative margin)
to allow for nested columns. (See http://getbootstrap.com/css/#grid-nesting) 
In the Dashboard theme, we are adding additional padding to .row, which breaks that system.

I just ran into this setting on a
dashboard-style demo, where I could not figure out why my nested columns
keep getting more and more indented.

I believe that this line of css was added to fix the search screen for
the dashboard theme. But a better fix is to use the "content" class
provided by AdminLTE. This is what I have done on my own demo, and it is
working well. It is also what the AdminLTE site does (though it is oddly
undocumented): https://almsaeedstudio.com/themes/AdminLTE/index2.html

### DIFF
--- a/app/themes/dashboard/ui/app/root/root.html
+++ b/app/themes/dashboard/ui/app/root/root.html
@@ -61,7 +61,7 @@
 
   <div class="content-wrapper" style="min-height:800px">
     <message-board msg="rootCtrl.messageBoardService.message()"></message-board>
-    <div ng-hide="rootCtrl.messageBoardService.message()">
+    <div ng-hide="rootCtrl.messageBoardService.message()" class="content">
       <ui-view></ui-view>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


grtjn:
This might have been fixed in the cards-themes branch. Can you check and change the pr to merge against that branch?

patrickmcelwee:
OK - not fixed there yet, but I changed the PR to point against the cards-theme branch

grtjn:
Merged..

@jenbreese our internal CI will have run with this change by tomorrow, could you take a look to see if the dashboard theme still looks right to your opinion?

jenbreese:
@grtjn and @patrickmcelwee, looks good to me on CI. thank you for the fix @patrickmcelwee!

PR #447, opened 2017-01-31
Create prod.properties and dev.properties placeholder files

Audarth:
Resolves SW-5 from our internal jira


grtjn:
We decided to go for the gulp init-prod approach instead. I'll be pushing a small fix for that later today, since it got broken because of Roxy changes..

PR #448, opened 2017-01-31
Make front end more fault tolerant to the session being lost

Audarth:
Resolves SW-6 from our internal jira

Added logic to the error interceptor so that on a 401 unauthorized, the user is redirected to the login screen.

### DIFF
--- a/app/templates/ui/app/error/errorInterceptor.service.js
+++ b/app/templates/ui/app/error/errorInterceptor.service.js
@@ -8,12 +8,14 @@
     }]);
 
   ErrorInterceptor.$inject = ['$q', '$injector', 'ngToast'];
+
   function ErrorInterceptor($q, $injector, toast) {
     var errorInterceptor = {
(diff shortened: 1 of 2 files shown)


app/templates/ui/app/error/errorInterceptor.service.js
```diff
@@ -8,24 +8,43 @@
...
+            msg = {
+              title: rejection.data.errorResponse.status,
+              body: rejection.data.errorResponse.message
+            };
+          } else {
+            msg = {
+              title: (rejection.status === 401) ? 'Unauthorised' : 'Internal Server Error',
+              body: ''
+            };
+          }
```
grtjn: Regarding the entire if above, isn't it always title Unauthorized? You only enter this branch if rejection.status === 401..

grtjn:
@Audarth We are merging this now to move forward. If you have further changes, please just open a new PR to pile on top of this..

PR #485, opened 2017-03-08
getAuthenticatedStatus needs to always return promise

### LINKED ISSUE #480: userService.getUser fails if not logged in

patrickmcelwee:
If the _isAuthenticated memo was already set, it was returned a bald
value, without wrapping in a promise

Also, I found and threw the error that wasn't showing up - which would
have helped debug this issue.

Fixes #480

### DIFF
--- a/app/templates/ui/app/login/login.service.js
+++ b/app/templates/ui/app/login/login.service.js
@@ -175,21 +175,12 @@
       }
 
       if (routeIsProtected(next.name)) {
-        var auth = service.getAuthenticatedStatus();
-
-        if (angular.isFunction(auth.then)) {
-          auth.then(function() {
-            if (!service.isAuthenticated()) {
-              //this does NOT block requests in a timely fashion...
-              blockRoute(event, next, nextParams);
-            }
-          });
-        }
-        else {
-          if (!auth) {
+        service.getAuthenticatedStatus().then(function() {
+          if (!service.isAuthenticated()) {
+            //this does NOT block requests in a timely fashion...
             blockRoute(event, next, nextParams);
           }
-        }
+        });
 
       }
     });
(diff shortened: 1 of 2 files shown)


joemfb:
I think i just ran into this in a work session with a prospect.

grtjn:
Thanks for the PR, super helpful!

And it looks very simple! I had come up with something more complex, like further down in that file, where a promise is created first.

By the way, now you have opened a PR for this anyhow. Something very similar is also with userService.getUser itself.

Also, it is easy to make these function return promises, but not sure if other code that relies on these will be expecting promises. Can you scan through the code to double-check?

patrickmcelwee:
Yeah, the symptom I ran into is that it would never prompt for login when going to any route except '/'.

I'll try to look at those questions by the end of the day, Geert. I also might look for other places that we don't throw errors from promises - since that would have saved us a good bit of time pinpointing this error.

grtjn:
That would be great, thanks!

patrickmcelwee:
My grep through the code indicates that consumers won't break because this is now always a promise. But we probably could simplify this bit of code that actually was implicitly checking whether it getAuthenticatedStatus returned a promise: https://github.com/marklogic/slush-marklogic-node/blob/master/app/templates/ui/app/login/login.service.js#L178-L190

grtjn:
Did you take a look at userService.getUser yet by any chance? It has a similar issue of not always returning a promise, I thought..

patrickmcelwee:
I think that is true. I just did the same kind of wrapping there (added a commit).

There is only one consumer of getUser in the code now, and it can handle a promise or not without changes ... but this is probably a good change to make things easier on future consumers.

grtjn:
Merging now! (sorry for the delay)

Issue #490, opened 2017-03-21
Remove the blue LESS files


jenbreese:
We need to remove the blue variables and blue less file since it is no longer brand compliant. New color schemes are available in the new to-be-announced BitBucket repo.

grtjn:
Thanks @jenbreese!

PR #491, opened 2017-03-21
fixes for #490 which is to remove the blue less files

jenbreese:
We need to remove the Blue color variant to more closely align with our branding. There are now more color options in the Bit Bucket repo.


grtjn:
Bit late, but never the less thanks @jenbreese!

PR #499, opened 2017-04-27
437 integrate socket.io in middle tier

janmichaelyu:


grtjn:
Looks very promissing, thanks! I need to run a sanity check before merging though..

PR #500, opened 2017-06-01
Optimize the similar extension

ryanjdew:
Have had the performance of the similar REST extension bite me in the past since it does a for loop over all the similar documents returned. Enhanced it to resolve the URIs straight from the indexes.


grtjn:
Awesome, and so simple!

PR #550, opened 2018-03-29
Removing unused switchLogin code

patrickmcelwee:
It also refers to an '/api/user/switch' middle-tier route that doesn't
exist.

I can see how this code could be helpful, and might actually be in use in some demos, but it seems to rely on a middle-tier route that is not in the project. I suggest removing the dead code, as in this PR.

Alternatively, the middle-tier it anticipates could be added, after vetting that it doesn't raise security concerns.

### DIFF
--- a/app/templates/ui/app/login/login.service.js
+++ b/app/templates/ui/app/login/login.service.js
@@ -75,15 +75,6 @@
       }, failLogin);
     }
 
-    function switchLogin(username) {
-      return $http.post('/api/user/switch', {
-        'username': username
-      }).then(function(response) {
-        loginSuccess(response);
-        return response;
-      }, failLogin);
-    }
-
     function loginPrompt() {
       var d = $q.defer();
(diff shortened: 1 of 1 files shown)


grtjn:
I think i simply forgot to add that middle-tier route. It is generic code based on the assumption that the demo users share a pwd. I can easily share it.

It is used in a lot of my demos, as it makes demoing real easy. You login once with a srrong pwd, and only after that you can use the switch endpoint.

It really is for demo purposes, but it would save me a lot of trouble if most code would be available already. I dont think it is a big security concern. We could wrap it in a hidden option, and disable it by default if you like.

grtjn:
Comments or toggles, both could work..

Present pwd? No. Thinking about it now makes me realize it must have been taking pwd from the session in middle-tier. That might mean it would not work with digest auth (in which we should not be saving pwd in the session)..

